PR #11, opened 2021-05-25
nextcloud: better validatation and progess tracking

gsanchietti:
Implement validation an progress protocol


gsanchietti:
Replaced by #14

PR #12, opened 2021-05-25
User interface

andre8244:
UI for NethServer 8


core/ui/public/config/config.production.js
```diff
@@ -0,0 +1,9 @@
+window.CONFIG = {
+  PRODUCT_NAME: "NethServer 8",
+  DOCS_URL: "https://www.nethserver.org/documentation/",
+  COMPANY_NAME: "Nethesis",
+  API_ENDPOINT: window.location.hostname + window.location.pathname + "api",
```
DavidePrincipi, line 5: Can we add `window.location.port` to preserve the port number?

It is useful if the port 443 is reached through a PF/SSH Tunnel and the port number needs to be preserved.

DavidePrincipi (approved):

PR #18, opened 2021-05-31
Add Loki module

Amygos:
Loki is a log aggregation system, main project site: https://grafana.com/oss/loki/
This PR adds two new components:
* `logcli`, a cli tool for query the Loki server for logs
* `loki`, the server that will receive, and store the logs

As default, the Loki server will listen on the IP address of the node's VPN interface, using the default fixed port `3100`.
The agents `promtail` and the clients `logcli` will assume that the default Loki server to use, is installed on the cluster leader node.

What is not present in this PR and is left for further work:
* Loki server port reservation
* HTTPS certificate and path reservation in the main host, (currently the `traefik` module don't allow this)
* Authentication for sending and querying logs
* A way to make visible at cluster level the default configurations like: Loki's URL and credentials

### DIFF
--- a/.github/workflows/loki.yml
+++ b/.github/workflows/loki.yml
@@ -1,0 +1,34 @@
+name: loki
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'loki/**'
+  pull_request:
+    types: ['opened', 'synchronize']
+    paths:
+      - 'loki/**'
+
+jobs:
+  push_image:
+    env:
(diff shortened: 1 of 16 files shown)


DavidePrincipi (commented):
- Can we use the host Traefik instance instead of running another one?
- Add .github/workflow/loki.yml
- Temp dir cleanup

DavidePrincipi (approved):

PR #32, opened 2021-06-23
ldapproxy. Add set-backend action

DavidePrincipi:
- Start the Nginx proxy after a backend is set.
- Include LDAP schema information.
- The set-backend action is designed for account providers:
  when an account provider is added to a node it has to be
  contextually registered as LDAP backend.
- Added example invocations in Samba README.md


gsanchietti (approved):
I've read the README and the flow seems good to me.
I have no idea about the implementation :see_no_evil:

PR #36, opened 2021-06-30
Authorize Task submission with api-server

DavidePrincipi:
Direct task submission in a Redis queue works if the caller is allowed to run LPUSH. This is true for the `cluster` agent only. As result, only the `cluster` agent is allowed to submit tasks for other agents/modules.

To overcome this limitation we can submit Tasks through Api-server. We can leverage its role based authorization system to grant fine-grained permissions on actions. Provider modules can expose a public API, consumed by other modules.

Major changes:
- Added clear-text http route to api-server backend, available from 127.0.0.1 and cluster VPN only
- Specify `parent` task id in Task submission
- New `org.nethserver.authorizations` image label. When a module is instantiated required roles are assigned to the module agent.

      org.nethserver.authorizations = traefik@node:routeadm

- Removed actions execution from `create-module` output. Partially reverts PR #29
- Reworked join-cluster to use the new `agent.tasks` Python module to send the `add-node` action to the cluster leader


gsanchietti (commented):
It's seems good, but it's too complex for me :man_shrugging:

gsanchietti (approved):

Amygos (approved):

PR #39, opened 2021-07-05
Configure ldapproxy and VPN after Samba DC provisioning

DavidePrincipi:
- Samba DCs need to talk each other through the VPN for replication: this PR adds to the core API the `add-route` and the `remove-route` actions, and to the node API the `update-routes` action.
- Ldapproxy is configured _after_ the DC is provisioned, to avoid conflicts with the installation order: provisioning order might differ.
- The provisioning procedure (join/new domain) is decided automatically: if a DC with the wanted realm/domain already exists a join is performed. 
- The `agent.tasks` module has been extended to support `redis://` endpoint. This is designed for actions running under the cluster-agent, as they have enough privileges to bypass the `api-server`. Bear in mind that when a task is submitted through the api-server, it is audited.


gsanchietti (approved):
:man_shrugging:

DavidePrincipi:
I'm going to split this PR in more parts

PR #41, opened 2021-07-07
dokuwiki: configure traefik route

gsanchietti:
Use new agent API so the UI do not need to call traefik.

TODO:
- [x] use JSON schema validation
- [x] update README


gsanchietti:
Manually merged

PR #48, opened 2021-07-14
Reconfigure VPN with update-routes action

DavidePrincipi:
Implement a basic API to route custom destinations through the cluster VPN and reconfigure VPN routes when a new node is added to the cluster.
- ~~`add-route`: add a destination (in CIDR notation) for the given node_id~~ moved to another PR
- ~~`remove-route`~~ moved to another PR
- `update-routes`: read the VPN configuration from Redis and apply it to the current node. Helper command `update-routes` is also in the bundle for manual invocation.

To implement the above API a new set of functions has been added: the `agent.tasks` module was extended with new tasks invocation styles:
- support `endpoint='redis://cluster-leader'`, really useful for cluster actions because it bypasses the api-server auditing
- `runp()` run a list of task in parallel
- `runp_nowait()` as above, but does not wait the tasks completion. Instead a list of task IDs is returned
- `runp_brief()` run the tasks list in parallel and handle the errors by writing to stderr. The number of failed tasks is returned.

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -167,6 +167,7 @@ cluster.grants.grant(rdb, action_clause="list-*", to_clause="reader", on_clause=
 cluster.grants.grant(rdb, action_clause="get-*",  to_clause="reader", on_clause='cluster')
 cluster.grants.grant(rdb, action_clause="show-*", to_clause="reader", on_clause='cluster')
 cluster.grants.grant(rdb, action_clause="read-*", to_clause="reader", on_clause='cluster')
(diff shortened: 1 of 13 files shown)


core/install.sh
```diff
@@ -26,7 +26,7 @@ source /etc/os-release
 
 echo "Install dependencies:"
 if [[ ${ID} == "fedora" ]]; then
-    dnf install -y wireguard-tools podman jq
+    dnf install -y wireguard-tools podman jq unzip wget
```
gsanchietti, line 29: Please add `jq unzip wget` also on line 43 for debian

PR #61, opened 2021-07-26
Store endpoint in IP format

DavidePrincipi:
If a VPN endpoint is given in host name form, resolve it to IP address with
DNS.

The DNS service might be not reachable during system boot. Configuring
the WireGuard device with a plain IP address ensures the interface is
started despite DNS failures.

If the node IP address changes over the time it is better to not set the
endpoint at all.

### DIFF
--- a/core/imageroot/usr/local/nethserver/agent/python/agent/__init__.py
+++ b/core/imageroot/usr/local/nethserver/agent/python/agent/__init__.py
@@ -212,15 +212,21 @@ def sanitize_endpoint(endpoint):
     if len(endpoint) == 0:
         return endpoint # empty string is allowed: it means there is no public VPN endpoint at all.
 
+    if ':' in endpoint:
+        nhost, nport = endpoint.split(':')
+    else:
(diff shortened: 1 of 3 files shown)


DavidePrincipi:
curl https://raw.githubusercontent.com/NethServer/ns8-scratchpad/ipendpoint/core/install.sh | bash -s ipendpoint core

gsanchietti (approved):
:see_no_evil:

DavidePrincipi:
From wg(8) manpage

> `WG_ENDPOINT_RESOLUTION_RETRIES`
>  If set to an integer or to infinity, DNS resolution for each peer's endpoint will be retried that many times for non-permanent errors, with an increasing delay between  re‐tries. If unset, the default is 15 retries.


Both Debian and Fedora set that value to `infinity`. Unless the DNS record is completely missing, a temporary failure should be already handled.

    Environment=WG_ENDPOINT_RESOLUTION_RETRIES=infinity

PR #65, opened 2021-07-28
Nextcloud,refactor, again

gsanchietti:


gsanchietti:
I'm going to put this on waiting list for a while, we need to focus on the core.

PR #77, opened 2021-09-14
install: set default password

gsanchietti:
Allow login to the UI before cluster initialization.

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -179,6 +179,9 @@ EOF
 echo "Install Traefik:"
 add-module traefik 1
 
+echo "Setting default admin password:"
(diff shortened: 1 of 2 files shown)


core/install.sh
```diff
@@ -179,6 +179,9 @@ EOF
 echo "Install Traefik:"
 add-module traefik 1
 
+echo "Setting default admin password:"
+add-user --role owner --password Nethesis,1234 admin
```
DavidePrincipi, line 183: What about adding an hook to override it?

```suggestion
add-user --role owner --password "${ADMIN_PASSWORD:-Nethesis,1234}" admin
```

gsanchietti: I like it

PR #88, opened 2021-09-20
core: sync ACLs

gsanchietti:
Synchronize redis ACLs from leader to worker nodes.
The synchronization is fired on:
- change-user-password
- add-user
- remove-user
- alter-user
- add-module

Above actions will save the ACLs to `cluster/acls` key and fire the `cluster/event/acl-changed` event.
All nodes will have a new service `enventsgw@node` to watch for the above event.
On `cluster/event/acl-changed` event, workers will execyte `/usr/local/bin/acl-load` (the command has no effect on leader node).
cute `/usr/local/bin/acl-load` when `acl-load` event has been received.

The following ACL users are not syncronized:
- cluster
- api-server
- default

Note: on remove-module action, the instance ACL is not removed so no event is fired.

See also https://trello.com/c/35M99jUh/61-redis-read-only-replicas


DavidePrincipi (commented):
ACLs have to be synced also
- in cluster/add-node/50update, when the node account is created
- at the end of cluster/join-cluster action: the worker node needs to apply the ACL after the Redis DB replica has successfully started

gsanchietti:
>     * at the end of cluster/join-cluster action: the worker node needs to apply the ACL after the Redis DB replica has successfully started

The joining node has already the synchronized key but it needs to execute the `load-acl` command.




>     * whenever `redis.service` starts (?)

The worker nodes should execute the `load-acl` whenever a synchronization on `cluster/acls` key occurs.
Maybe we can use [keyspace notifications](https://redis.io/topics/notifications).

core/imageroot/usr/local/bin/acl-load
```diff
@@ -0,0 +1,49 @@
...
+        user = acl.split(" ",2)[1]
+        # replace existing user
+        rdb.acl_deluser(user)
+        rdb.execute_command(f'ACL set{acl}')
+
+    rdb.acl_save()
+    print("ACLs loaded")
+else:
+    print("This command should run only on worker node", file=sys.stderr)
+    sys.exit(1)
```
DavidePrincipi, line 49:
```suggestion
    print("This is a leader node, nothing to do", file=sys.stderr)
    sys.exit(0)
```

gsanchietti: I prefer to have a clear statement: the command failed.

Is like calling a program without required parameters: it exists with non-zero code

PR #90, opened 2021-09-22
Fix agent.read_envfile()

DavidePrincipi:
Split on "=" can occur only one time


core/imageroot/usr/local/nethserver/agent/python/agent/__init__.py
```diff
@@ -76,7 +76,7 @@ def read_envfile(file_path):
     for line in fo.readlines():
         lineno =+ 1
         try:
-            variable, value = line.split("=")
+            variable, value = line.split("=", 1)
```
DavidePrincipi, line 79:
```suggestion
            variable, value = line.strip().split("=", 1)
```

gsanchietti (approved):

PR #93, opened 2021-10-06
Fix Traefik systemd unit deps loop

DavidePrincipi:
Running Traefik `Before=certificate-exporter.path` is necessary to create the working directory before the certificate-exporter.service starts

However that Before= directive creates a startup dependency ordering loop because .path units are scheduled to start before the .service ones.

The PR 
- attempts to fix the unit startup ordering issue
- considers a missing `acme.json` (or any of its path components) as non-fatal for the `export-certificate` script

### DIFF
--- a/traefik/imageroot/bin/export-certificate
+++ b/traefik/imageroot/bin/export-certificate
@@ -33,32 +33,32 @@ import os
 import json
 import agent
 import os.path
+import sys
 
-def _read_and_set():
-    module_id = os.environ['MODULE_ID']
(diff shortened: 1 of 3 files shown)


gsanchietti (commented):
I really do not get why the export-certificate script as been refactored again. 

About systemd configuration: original implementation worked for Debian but it has been modified for Fedora. It's a bit strange that systemd behaves in a different way in Debian and Fedore. Are you sure this change is really needed?

If you think it's necessary, please go ahead.

PR #95, opened 2021-10-11
Do not touch Redis */environment keys

DavidePrincipi:
PR goals:

- An agent ignores the values of its */environment Redis key: it only writes on it, no reads are performed.
- The agent is the only responsible for updating its */environment key. The core no longer writes the key.
- The */environment key fields are updated by the agent during each action execution

To achieve the above goals:

- agent.go prepareActionEnvironment() function reads the environment from the local filesystem instead of the Redis */environment key
- the node agent creates the initial environment file for the new module agent. The environment is created by the cluster agent and passed as input to the [node/add-module action](https://github.com/NethServer/ns8-scratchpad/blob/c79ad8b1891e780018962e4d223073eb1c6ca554/core/add-module-input.md#environment)
- similarly, node and cluster agent environments are initialized by install.sh, create-cluster and join-cluster actions

### DIFF
--- a/core/agent/README.md
+++ b/core/agent/README.md
@@ -71,7 +71,10 @@ When all steps are executed successfully the action is considered `completed`. T
 - `<agent-id>/task/<task-id>/output`, collects data sent to FD 1 by all steps
 - `<agent-id>/task/<task-id>/error`, collects data sent to FD 2 by all steps
 - `<agent-id>/task/<task-id>/exit_code`, 0 if all steps were successful, otherwise the exit code of the failing step
-- `<agent-id>/environment`, additional environent variables passed to the action steps. The values are persisted to Redis if the action is completed
+- `<agent-id>/environment`, copy of additional environment variables passed
+  to the action steps. The values are stored in the local module
+  filesystem and copied to Redis if the action has completed successfully
+  or the `dump-env` command was called.
 
 The `validation-failed` status must be set manually with the `set-status` command, described below. When the `validation-failed`
 status is set, no more steps are invoked and the action exit code is set to 10 unless another non-zero exit code is returned by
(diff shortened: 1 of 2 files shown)


core/agent/agent.go
```diff
@@ -64,17 +64,19 @@ var taskExpireDuration = 24 * time.Hour
...
-	for key, value := range redisHash {
-		env = append(env, key+"="+value)
+	for _, line := range strings.Split(string(content), "\n") {
+		if line == "" {
+			continue
+		}
+		if strings.Contains(line, "=") == false {
+			continue
+		}
+		env = append(env, line)
```
gsanchietti, line 79: This doesn't correctly support env files like:
```
VAR1="value1"
```

core/imageroot/var/lib/nethserver/cluster/actions/add-node/50update
```diff
@@ -85,13 +85,6 @@ agent.assert_exp(rdb.hset(f'node/{node_id}/vpn', mapping={
     "listen_port": str(listen_port),
 }) >= 0)
 
-#
-# 1/d. Initialize the node agent environment
-#
-agent.assert_exp(rdb.hset(f'node/{node_id}/environment', mapping={
-    "NODE_ID": node_id
-}) >= 0)
-
```
gsanchietti, line 94: Why this check has been removed? Redis should already contain the information.

gsanchietti: My bad, we should not write anymore on redis env, this is why it has been removed: it's ok.

core/agent/agent.go
gsanchietti: Let's keep this implementation unchanged, it's the simplest one.

DavidePrincipi:
Rebasing to fix merge conflicts...

PR #114, opened 2021-10-29
Protect runagent from special shell chars

DavidePrincipi:
The "source" of environment files is fragile: if they contain special
shell chars, like "&" the parsing fails.

Read environment files and export the environment variables
without importing them in Bash, thus special chars are not considered.

Changed `agent.read_envfile()`: values of variables are stored in raw form, no shell-like special chars escaping and transformations are needed.

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
+++ b/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
@@ -73,16 +73,17 @@ def read_envfile(file_path):
     lineno = 0
     env = {}
     for line in fo.readlines():
-        lineno =+ 1
+        lineno += 1
         try:
-            if line[0] == '#':
-                continue # skip comments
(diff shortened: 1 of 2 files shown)


core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -75,6 +75,8 @@ def read_envfile(file_path):
     for line in fo.readlines():
         lineno =+ 1
         try:
+            if line[0] == '#':
+                continue # skip comments
```
gsanchietti, line 79:
```suggestion
            if line[0] == '#' or not line.strip():
                continue # skip comments and empty lines
```

gsanchietti (commented):
PR looks good, but I propose just a small fix on the parsing function.

gsanchietti (approved):

PR #118, opened 2021-11-08
install: support Fedora Server 35

gsanchietti:
Fedora Server 35 needs to build many python dependencies.

pip needs to build some binary dependencies because there is no Python [wheel](https://packaging.python.org/tutorials/installing-packages/#source-distributions-vs-wheels) available for the following packages:
- cchardet
- cffi
- yarl
- multidict
- pycares
- psutils
- hiredis

Removing the specific version from `pythonreq.txt` could fix the problem for all packages except `hiredis` and `psutils`

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -28,6 +28,10 @@ echo "Install dependencies:"
 if [[ ${ID} == "fedora" ]]; then
     dnf install -y wireguard-tools podman jq openssl
     systemctl disable --now firewalld || :
+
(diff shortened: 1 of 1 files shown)


gsanchietti:
Needed for pip to build:
- cchardet
- cffi
- yarl

gsanchietti:
Before merge, let's wait upstream releases new wheels for Python 3.10

gsanchietti:
Let's close this one, the support should be almost automatic when Python wheels will provides binaries for Python 3.10

PR #121, opened 2021-11-11
Store more artifacts in the core build cache

DavidePrincipi:
- Save Golang modules and build cache
- Save buildah overlay-images directory

The cache expires every Sunday, and is invalidated if go.sum go.mod yarn.lock and build-image.sh are modified.

For instance, this is a cache key:

    Cache restored from key: corecache-2021W45-715a07faf8426d54f9d1d207433c1cbff9c762f8cb1a73e6784442eb2f9b736b

See also https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -19,14 +19,21 @@ jobs:
     name: 'Build ${{ github.workflow }}'
     runs-on: ubuntu-20.04
     steps:
+      - id: info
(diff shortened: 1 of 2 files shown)


gsanchietti (approved):
:crossed_fingers:

PR #131, opened 2021-11-23
MariaDB module

stephdl:
The goal of the PR is to install phpmyadmin and mariadb on ns8, two containers inside a pod, phpmyadmin-app and mariadb-app. All secrets are mounted by the `--secrets` command of podman that are read directly by the container (for instance one example for mariadb root password `-e MARIADB_ROOT_PASSWORD_FILE=/run/secrets/root.secret`

there is  volumes created to store custom configurations or mysql data

```
DRIVER      VOLUME NAME
local       config.user.inc.php    # phpmyadmin custom configuration
local       mysql-conf.d             # mariadb specific configuration
local       mysql-data                # mariadb data
local       sql                            # mount inside any .sql file that will be read only at the first boot of the mariadb container
```

WIP area

- ~~UI~~
- ~~/webpath validation~~
- enhance security if I have broken the rule

to install the pod 

```
git clone the repo
cd mariadb/
build-image.sh
```

install the module

`add-module ghcr.io/nethserver/mariadb:latest 1
`
configure the module 
```
api-cli run configure-module --agent module/mariadb1 --data - <<EOF
{ 
"path": "/phpmyadmin", 
"http2https": true,
}
EOF
```
The external port for the mysql client is random (> 20000) like the port for the web application container 

the phpmyadmin is located at `http://hostname/path`, you can find the link inside the UI
 
you can query for the UI by

`[root@fedora ~]# api-cli run get-configuration --agent module/mariadb12 --data null`

it is possible to use directly mysql inside the container

```
ssh mariadb1@localhost
podman exec -ti mariadb-app mysql
```

Screenshot (1)
Screenshot


stephdl:
obsoleted by https://github.com/NethServer/ns8-mariadb/pull/1

PR #145, opened 2021-12-17
CentOS stream 9 support

gsanchietti:
- Support CS9 inside install.sh
- Improve promtail configuration for CS9

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -25,10 +25,10 @@ set -e
 source /etc/os-release
 
 echo "Install dependencies:"
-if [[ ${ID} == "fedora" || ${PLATFORM_ID} == "platform:el9" ]]; then
+if [[ "${PLATFORM_ID}" == "platform:f34" || "${PLATFORM_ID}" == "platform:el9" ]]; then
     dnf install -y wireguard-tools podman jq openssl
     systemctl disable --now firewalld || :
-elif [[ ${ID} == "debian" ]]; then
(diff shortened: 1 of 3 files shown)


DavidePrincipi:
What about enabling the "persistent storage" directory?

From systemd-journald man page:

       The journal service stores log data either persistently below /var/log/journal or in
       a volatile way below /run/log/journal/ (in the latter case it is lost at reboot). By
       default, log data is stored persistently if /var/log/journal/ exists during boot,
       with an implicit fallback to volatile storage otherwise. Use Storage= in
       journald.conf(5) to configure where log data is placed, independently of the
       existence of /var/log/journal/.

       On systems where /var/log/journal/ does not exist yet but where persistent logging is
       desired (and the default journald.conf is used), it is sufficient to create the
       directory, and ensure it has the correct access modes and ownership:

           mkdir -p /var/log/journal
           systemd-tmpfiles --create --prefix /var/log/journal

gsanchietti:
> What about enabling the "persistent storage" directory?

The persistent storage is already done by Loki server, there is no need to have another persistent storage system.

What do you think @Amygos ?

Amygos:
I agree with @gsanchietti and I prefer to don't change the behavior of the underlying operating system.

DavidePrincipi (approved):
Please wait a while (until CentOS Stream 9 images are available from cloud providers) before merging this PR.

Amygos (approved):

PR #156, opened 2022-01-31
uninstall: remove all rootfull services

gsanchietti:
Make sure also promtail1 has been uninstalled.

Without this fix, after the uninstall, `promtail1` service was still in place.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/uninstall.sh
+++ b/core/imageroot/var/lib/nethserver/node/uninstall.sh
@@ -37,7 +37,8 @@ for modulehome in /var/lib/nethserver/*; do
     fi
     moduleid=$(basename $modulehome)
     echo "Deleting rootfull module ${moduleid}..."
-    systemctl disable --now eventsgw@${moduleid} agent@${moduleid} && rm -rf "${modulehome}"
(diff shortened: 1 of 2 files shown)


core/imageroot/var/lib/nethserver/node/uninstall.sh
```diff
@@ -37,7 +37,8 @@ for modulehome in /var/lib/nethserver/*; do
     fi
     moduleid=$(basename $modulehome)
     echo "Deleting rootfull module ${moduleid}..."
-    systemctl disable --now eventsgw@${moduleid} agent@${moduleid} && rm -rf "${modulehome}"
+    systemctl disable --now eventsgw@${moduleid} agent@${moduleid} ${moduleid} && rm -rf "${modulehome}"
+    rm -f /etc/systemd/system/${moduleid}.service
```
DavidePrincipi, line 41: There can be more than one service started by a rootfull module. We could try to iterate over all units starting with ${moduleid}. For instance

    systemctl list-unit-files -q  -o json "${moduleid}*"  | jq -r .unit_file

DavidePrincipi: I tried to implement the cleanup in #185

PR #164, opened 2022-02-09
core: add GitHub Actions as supported platform

Amygos:
Supporting the installation on GitHub Actions runners based on Ubuntu
20.04, will allow the execution of the tests for the external module
without the need for external infrastructure.


gsanchietti (approved):
Is it so easy!? Cool!

Amygos:
> Is it so easy!? Cool!

The tricky part is that the script must be executed with sudo in this way:
```
 sudo -E PATH=$PATH HOME=/root bash install.sh
 ```
 and you must setup python 3.9 before the installation process:
 ```yaml
       - name: Setup Python
        uses: actions/setup-python@v2.3.1
        with:
          python-version: 3.9
```
This part will land on a PR to https://github.com/NethServer/ns8-github-actions

DavidePrincipi (approved):

PR #166, opened 2022-02-11
install.sh: restart journald

gsanchietti:
Sometimes, `journalctl` is not correctly displaying the messages.

This is has been tested on a Debian 11 running on Digital Ocean:
```
# journalctl 
No journal files were found.
-- No entries --
```

The fix:
```
# systemctl restart systemd-journalctl
```

Outcome:
```
# journalctl 
-- Journal begins at Fri 2022-02-11 13:15:51 UTC, ends at Fri 2022-02-11 13:15:51 UTC. --
Feb 11 13:15:51 debian-s-1vcpu-1gb-intel-lon1-01 systemd[1]: Stopping Journal Service...
Feb 11 13:15:51 debian-s-1vcpu-1gb-intel-lon1-01 systemd-journald[222]: Received SIGTERM from PID 1 (systemd).
Feb 11 13:15:51 debian-s-1vcpu-1gb-intel-lon1-01 systemd[1]: systemd-journald.service: Succeeded.
...
```

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -39,6 +39,11 @@ else
     exit 1
 fi
 
+if [[ $(journalctl --output cat -n 1) == "" ]]; then
+    echo "Journal is empty, attempting a systemd-journald restart:"
+    systemctl restart systemd-journald.service
+fi
+
 echo "Set kernel parameters:"
 sysctl -w net.ipv4.ip_unprivileged_port_start=23 -w user.max_user_namespaces=28633 -w net.ipv4.ip_forward=1 | tee /etc/sysctl.d/80-nethserver.conf
 
@@ -190,9 +195,6 @@ add-user --role owner --password "${ADMIN_PASSWORD:-Nethesis,1234}" admin
 
 echo "Enable the events gateway for the node agent:"
 systemctl enable --now eventsgw@node
-
-echo "Restart systemd-journald:"
-systemctl restart systemd-journald
 
 cat - <<EOF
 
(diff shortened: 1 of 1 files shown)


Amygos (approved):

core/install.sh
```diff
@@ -191,6 +191,9 @@ add-user --role owner --password "${ADMIN_PASSWORD:-Nethesis,1234}" admin
 echo "Enable the events gateway for the node agent:"
 systemctl enable --now eventsgw@node
 
+echo "Restart systemd-journald:"
+systemctl restart systemd-journald
+
```
DavidePrincipi, line 196: If the change is just a workaround to a Debian issue, can we put in at line L33?

gsanchietti: I've found the error only on Debian, but maybe Ubuntu for CI is also affected.
And maybe even CS9 on other cloud platforms can be affected.

I do not see any drawback on restarting journald. Do you think we could face any problem?

gsanchietti: > Am I wrong?

The installation log is not lost. You can see it using `journalctl -m`.

The restart is need to see the logs form now on. If you do not restart it, promtail will not be able to see coming messages.

> I've another concern. If the install script runs during the machine first boot (i.e. cloud-init) is it safe to restart journald at that point?

Why it shouldn't be? By the way, for now installation is happening outside cloud-init

Amygos (approved):

PR #175, opened 2022-02-18
Refactoring of journal `CONTAINER_TAG` field

Amygos:
- core. add default log tag to rootless modules
- Remove `--log-opt` option from rootless modules
- core, test: add rootless logs tests
- Unified use of `--log-opt` for rootfull modules
- core, test: add rootfull logs tests


gsanchietti (approved):
After merge, remember also to remove the tag from https://github.com/NethServer/ns8-kickstart/blob/main/imageroot/systemd/user/kickstart.service#L19

DavidePrincipi (approved):
:+1: very good!

PR #183, opened 2022-02-23
Fix race conditions during aioredis cleanup

DavidePrincipi:
Execute Python cancel-task handlers in a new context to suppress errors
during async clean up procedures.

The same approach was applied also to the aiohttp client.


gsanchietti (commented):
:man_shrugging:

gsanchietti:
Installation tested without errors

PR #186, opened 2022-02-25
Authorize action with pattern match

DavidePrincipi:
Change the api-server to enforce shell-like pattern matching rules when checking the action authorization.

Pattern matching saves the need of calculating and updating all possible
action names on the API/backend side. Maintaining a list of patterns is
easier than maintaining the full action list for the current exact-match
rule.

See also https://pkg.go.dev/path/filepath#Match


edospadoni (approved):

DavidePrincipi:
New ACLs format for a worker node

```
127.0.0.1:6379> SMEMBERS node/2/roles/owner
1) "*"
127.0.0.1:6379> SMEMBERS node/2/roles/reader
1) "show-*"
2) "list-*"
3) "read-*"
4) "get-*"
```

gsanchietti (approved):

Amygos (approved):

PR #192, opened 2022-03-07
Include only state and volumes in backups

DavidePrincipi:
To avoid issues with nested container mountpoints during restore/backup procedures, limit mounted paths to

- state/ 
- volumes/

Files outside of those paths must not constitute the module state and cannot be included in the state backup by any means.

This PR includes a refactor commit that removes the module-restore agent command (see commit 5dd6b144366aa75eae856982244f86c8b0a2320b).


core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -174,6 +174,41 @@ def run_restic(rdb, repository, repo_path, podman_args, restic_args, **kwargs):
...
+                    elif module_is_rootfull:
+                        dvolume = module_id + '-' + svolume
+                    else:
+                        dvolume = svolume
+
+                    volume_args[dvolume] = f"--volume={dvolume}:/srv/volumes/{svolume}"
+    else:
+        # TODO: As alternative default strategy,
+        #       we could mount already existing volumes
+        raise Exception("Not implemented")
```
gsanchietti, line 207: Can we cleanup this code?

```suggestion
        return list()
```

PR #194, opened 2022-03-08
Fix default HTTPS certificate

Amygos:
If the node's FQDN is used in the default selfsigned certificate, when a valid certificate for the node's FQDN is requested using the `set-certificate` action, `traefik` will not request a new certificate via Let's Encrypt, but will use the already present selfsigned certificate.
This is caused by the fact that in the system is already present a certificate for the requested domain (in the selfsigned certificate) and to avoid this problem the tld `.test` is used in the selfsigned certificate.
The `CN` of the selfsigned certificate will be `FQDN + .test`

### DIFF
--- a/traefik/imageroot/actions/create-module/20certificate
+++ b/traefik/imageroot/actions/create-module/20certificate
@@ -31,7 +31,7 @@ set -e
 exec 1>&2
 
 NAME="selfsigned"
-FQDN=$(hostname -f)
+FQDN="$(hostname -f).test"
 addresses=''
 
 # Do not override existing certs
--- a/core/tests/10_cluster_sanity/80_trefik_check_certificate_apis.robot
+++ b/core/tests/10_cluster_sanity/80_trefik_check_certificate_apis.robot
@@ -1,0 +1,12 @@
+*** Settings ***
+Library    RequestsLibrary
+Library    SSHLibrary
+
+*** Test Cases ***
+Set node FQDN certificate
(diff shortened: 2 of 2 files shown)


gsanchietti (commented):
It looks good to me.
Still, I remember that old Firefox versions did not permit to bypass the control of such certificate.
I need to test it a little bit.

gsanchietti (approved):
Tests are ok, both on Windows and Linux

traefik/imageroot/actions/create-module/20certificate
```diff
@@ -31,7 +31,7 @@ set -e
 exec 1>&2
 
 NAME="selfsigned"
-FQDN=$(hostname -f)
+FQDN="ns8.local"
```
DavidePrincipi, line 34: The .local suffix is reserved for other purposes. What about appending .test to the FQDN?

```suggestion
FQDN="$(hostname -f).test"
```

gsanchietti: The browser will complain, but as long as HSTS is not enable on that domain, it should not be a real problem.

DavidePrincipi: :raised_hands:

DavidePrincipi (approved):

traefik/imageroot/actions/create-module/20certificate
Amygos: In the end, I pick the solution proposed by @DavidePrincipi :)

PR #198, opened 2022-03-11
traefik: use fixed length `CN` in self signed cert

Amygos:
The `CN` of the self signed certificate was made by concatenating the
`FQDN` of the node with the `.test` tld, this can cause a failure during
the certificate generation if the `FQDN` is already at maximum allowed
length. This was the case with the GitHub Actions runner.

To solve this problem the value of `CN` will be:
* `host-<five random numbers>.ns8.test`


DavidePrincipi (approved):
:heart:

gsanchietti (approved):
:scream:

PR #200, opened 2022-03-15
Implement backup and restore

gsanchietti:
- Add state-include.conf
- Implement restore-module action


gsanchietti:
This PR just adds loki to backup to preserve the logs, if needed.
A future improvement could lead to promtail reconfiguration after loki restore.

PR #220, opened 2022-03-23
install.sh: install podman => 3.4.2

Amygos:
The version provided by Debian 11 of the `containers-common` package,
provided as dependence by `podman`, don't have the support for `log-tag`
an option that is needed by the logs system collector.

See: https://github.com/containers/common/commit/6a2f926c6498ce9f76304933354f297dd52302b5


gsanchietti (approved):
I was expecting it :)

This is ok for me, but I do not know how much can impact the updates of Debian itself.

DavidePrincipi (approved):
I'd prefer getting Podman from Debian testing repo, but it's ok for a temporary workaround!

DavidePrincipi (approved):

PR #237, opened 2022-04-12
Log page. Filter like journalctl -t

DavidePrincipi:
The SYSLOG_IDENTIFIER journal field is more flexible: it works with
applications that are not aware of containers (i.e. agents).


gsanchietti (commented):
I do not understand what changes this PR introduce.
As long as current implementation should not be changed in all modules, I'm ok with it.
Also I think  logs should be kept associated to machines using the hostname.

DavidePrincipi:
I updated the issue description with more information. I hope it is more clear now.

Amygos (approved):

gsanchietti (approved):
:man_shrugging: 

In @Amygos we trust

edospadoni (approved):

PR #239, opened 2022-04-13
Use pipeline mode in redis-exec command

DavidePrincipi:
Optimize the multiple commands execution in a single pipeline.

### DIFF
--- a/core/imageroot/usr/local/agent/bin/redis-exec
+++ b/core/imageroot/usr/local/agent/bin/redis-exec
@@ -24,14 +24,19 @@ import sys
 import agent
 import csv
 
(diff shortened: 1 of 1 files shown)


DavidePrincipi:
Test suite passed here https://github.com/NethServer/ns8-core/runs/6006136984

Amygos:
Can you show an example?

DavidePrincipi:
The command input and output formats do not change at all. However a Redis command pipeline is sent instead of a command/response loop, when the commands are read from standard input. E.g.

    redis-exec <<EOF
    SET a b
    SET c d
    EOF

Amygos (approved):

PR #240, opened 2022-04-22
Drop agent package dependency in cluster.userdomains

DavidePrincipi:
Existing similar functions receive a Redis opened connection handle
as mandatory argument. The package should not open new connections by
itself.


DavidePrincipi:
Existing function calls seem to always pass `rdb`: https://github.com/search?q=org%3ANethServer+list_domains&type=code

gsanchietti:
I'd prefer to have a simpler interface for developers.
What is the reason to drop current implementation?

PR #245, opened 2022-05-15
Update build-image.sh

aereaco:
set supervised to no in cat >etc/redis.conf


DavidePrincipi (changes requested):
According to https://github.com/redis/redis/blob/e6f67092f8d4d81761a60c46011d1ff1dc3c2628/redis.conf#L274-L275 the default value is already `no` and the PR is useless.

I'm closing the PR.

PR #251, opened 2022-05-26
Fix api-cli auto completion

DavidePrincipi:
Since #186 the `api-cli` Bash auto completion is broken. This PR aims to fix it. Refs fb1e605e8f6a440aa03fecbc761f8dbbe4ba2625

The completion token is passed by the Bash function to `api-cli list-actions [prefix]`, as the search `prefix`.

The prefix is used to generate the correct completion list. Only when a full AGENT_ID is completed, the full action list is retrieved with a `list-actions` action call.


gsanchietti (commented):
I have no idea, I trust you :)

PR #264, opened 2022-07-12
core. api-server: 2fa first stub

edospadoni:
Added 2FA using Google Authenticator or similar apps.

https://trello.com/c/eNbtryd7/201-2fa-authentication


edospadoni:
Replace by https://github.com/NethServer/ns8-core/pull/267

PR #273, opened 2022-07-26
Add docs note for images in Podman local storage

DavidePrincipi:
During development, some local Podman images might override remote ones, producing unexpected behavior.

The typical case is when a `:latest` image tag is used to install/update a module instance that was already downloaded.


Amygos (approved):

gsanchietti (approved):
:+1:

stephdl (approved):

PR #285, opened 2022-08-05
Clean up old module and core images

DavidePrincipi:
- When a module is updated, remove the old image from the Podman storage
- When the core is updated, remove the old core images from the Podman
storage

Removing the old images frees up disk space and forces the image
complete download, which is important during development to avoid
unexpected install of stale images, especially in worker nodes.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/actions/update-core/95cleanup_images
+++ b/core/imageroot/var/lib/nethserver/node/actions/update-core/95cleanup_images
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+#
+# Copyright (C) 2022 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+exec 1>&2
+set -e
+
+# Clean up local images that are no longer required by the new core
+# version
+
+export LC_ALL=C # required by sort for portability
+
+source /etc/nethserver/core.env
+
+# Find the images that are no longer required
+diff <(
+    echo "${IMAGE_ID}"
+    podman inspect -f json "${IMAGE_ID:-none}" | \
+            jq -r '.[0].Config.Labels."org.nethserver.images" | split(" ") | join("\n")' | \
+            sort -c
+) <(
+    echo "${PREV_IMAGE_ID}"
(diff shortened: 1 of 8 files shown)


core/imageroot/var/lib/nethserver/node/actions/update-core/95cleanup_images
```diff
@@ -0,0 +1,37 @@
...
+    echo "${PREV_IMAGE_ID}"
+    podman inspect -f json "${PREV_IMAGE_ID:-none}" | \
+            jq -r '.[0].Config.Labels."org.nethserver.images" | split(" ") | join("\n")' | \
+            sort
+) <(
+    echo "${IMAGE_ID}"
+    podman inspect -f json "${IMAGE_ID:-none}" | \
+            jq -r '.[0].Config.Labels."org.nethserver.images" | split(" ") | join("\n")' | \
+            sort
+) | sed -n '/< / s/< //p' | \
```
gsanchietti, line 36: I'm not really happy with the `sed` part, maybe it's better to use a safer diff output?

Still, I'm not sure it worth the change.
```suggestion
diff -u <(
    echo "${PREV_IMAGE_ID}"
    podman inspect -f json "${PREV_IMAGE_ID:-none}" | \
            jq -r '.[0].Config.Labels."org.nethserver.images" | split(" ") | join("\n")' | \
            sort
) <(
    echo "${IMAGE_ID}"
    podman inspect -f json "${IMAGE_ID:-none}" | \
            jq -r '.[0].Config.Labels."org.nethserver.images" | split(" ") | join("\n")' | \
            sort
) | grep -e '^-' | cut -d '-' -f2
```

gsanchietti (approved):

core/imageroot/var/lib/nethserver/node/actions/update-core/95cleanup_images
DavidePrincipi: Ok to add the regexp anchor. About `-u` I prefer to use the normal diff format, like similar code here:
https://github.com/NethServer/ns8-core/blob/8efbdeb19f6de6145531142b4d5ac90a36a24fc7/core/imageroot/usr/local/agent/bin/install-coreimage#L49

PR #293, opened 2022-09-12
Add counters to list-user-domains API

DavidePrincipi:
Return the number of users and groups in the domain as cached from the last run of list-domain-groups/users.


gsanchietti (approved):
Ok, remember to add the new key inside the doc: https://github.com/NethServer/ns8-core/blob/main/docs/core/database.md

PR #295, opened 2022-09-15
core. api-server & ui: added timezone in logs page

edospadoni:
https://trello.com/c/vfKe3bfH/213-log-page-vs-timezone

### DIFF
--- a/core/api-server/socket/action.go
+++ b/core/api-server/socket/action.go
@@ -147,7 +147,6 @@ func Action(socketAction models.SocketAction, s *melody.Session, wg *sync.WaitGr
 		args = append(args, entity)
 
 		// define command
-		fmt.Println("TZ="+timezone+" /usr/local/bin/logcli", args)
 		cmd := exec.Command("/usr/local/bin/logcli", args...)
 
 		// add envs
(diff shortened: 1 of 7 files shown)


andre8244 (approved):
UI part LGTM

core/api-server/socket/action.go
```diff
@@ -137,8 +144,10 @@ func Action(socketAction models.SocketAction, s *melody.Session, wg *sync.WaitGr
 		args = append(args, entity)
 
 		// define command
-		fmt.Println("/usr/local/bin/logcli", args)
+		fmt.Println("TZ="+timezone+" /usr/local/bin/logcli", args)
 		cmd := exec.Command("/usr/local/bin/logcli", args...)
+		cmd.Env = os.Environ()
+		cmd.Env = append(cmd.Env, "TZ="+timezone)
```
Amygos, line 150: For me is better if the default behavior of the `logcli` command is preserved when no `--timezone` parameter is passed,
```suggestion
		cmd := exec.Command("/usr/local/bin/logcli", args...)
		if len(logsAction.TimeZone) > 0 {
			cmd.Env = os.Environ()
			cmd.Env = append(cmd.Env, "TZ="+timezone)
			fmt.Println("TZ="+timezone+" /usr/local/bin/logcli", args)
		} else {
			fmt.Println("/usr/local/bin/logcli", args)
		}
```

edospadoni: I like your idea, but I made a better implementation without using your suggestion. Let me know if you like it.

Amygos (approved):

PR #305, opened 2022-10-27
Terminate the user session before to remove the module

stephdl:
When you remove a rootless module, we have to terminate (kill)  all sessions relevant to the user before to remove the module with its associated user.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
+++ b/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
@@ -42,6 +42,9 @@ if [[ -z $uid ]]; then # Rootfull module
 else # Rootless module
     loginctl disable-linger "${mid}"
     systemctl stop "user@${uid}.service"
+    if pgrep -u "${mid}"; then
+        loginctl terminate-user "${mid}"
+    fi
 
     if [[ ${preserve} == "yes" ]]; then
         userdel "${mid}"
--- a/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
+++ b/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
@@ -42,9 +42,7 @@ if [[ -z $uid ]]; then # Rootfull module
 else # Rootless module
     loginctl disable-linger "${mid}"
     systemctl stop "user@${uid}.service"
-    if pgrep -u "${mid}"; then
-        loginctl terminate-user "${mid}"
-    fi
+    loginctl terminate-user "${mid}" &>/dev/null || :
 
     if [[ ${preserve} == "yes" ]]; then
         userdel "${mid}"


DavidePrincipi (commented):
Can we ignore this action exit code in the [caller action](https://github.com/NethServer/ns8-core/blob/main/core/imageroot/var/lib/nethserver/cluster/actions/remove-module/50update#L74
)?

From that line until the end of the file, let's remove agent.assert_exp (lines 74, 94, 97)

We could substitute them with warning messages... What do you think?

core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
```diff
@@ -42,6 +42,9 @@ if [[ -z $uid ]]; then # Rootfull module
 else # Rootless module
     loginctl disable-linger "${mid}"
     systemctl stop "user@${uid}.service"
+    if pgrep -u "${mid}"; then
```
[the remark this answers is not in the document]
DavidePrincipi: ...as alternative, rather than doing a test we'd better ignore the exit code of the next command.

I think it's even shorter!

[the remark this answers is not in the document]
stephdl:
```
[root@ns8loc2 ~]# loginctl user-status does_exist_user /dev/null 2>&1
Failed to look up user does_exist_user: No such process
[root@ns8loc2 ~]# echo $?
1
```
if the user does not exist or not logged we fail, I prefer to test if a processus exist for the user

:-?

core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
```diff
@@ -42,6 +42,9 @@ if [[ -z $uid ]]; then # Rootfull module
...
+        loginctl terminate-user "${mid}"
```
DavidePrincipi, line 46: Ignore non-existing user errors

```suggestion
        loginctl terminate-user "${mid}" &>/dev/null || :
```

core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
```diff
@@ -42,6 +42,9 @@ if [[ -z $uid ]]; then # Rootfull module
...
+    fi
```
DavidePrincipi, line 47: Ignore non-existing user error

```suggestion
    loginctl terminate-user "${mid}" &>/dev/null || :
```

stephdl: can we go to 

`loginctl terminate-user "${mid}" &>/dev/null || true`

DavidePrincipi: :shrug: They are equivalent, they both set the expression exit code to 0 overriding `set -e` behavior.

https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-_003a

stephdl: lets go to your proposal then.

DavidePrincipi (approved):

PR #307, opened 2022-10-28
Find the rootfull volumes and remove them

stephdl:
When we remove a rootfull container, the volumes are not destroyed even if asked

We cannot inspect the containers because at this point the container has been probably removed by the destroy-module action of the container. 

The possibility I see is to state that the documentation asks the volume must be named by  "module_id-something", therefore we could find the volumes and remove them.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
+++ b/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
@@ -38,6 +38,7 @@ if [[ -z $uid ]]; then # Rootfull module
     systemctl disable --now agent@"${mid}"
     if [[ ${preserve} == "no" ]]; then
         rm -rf "/var/lib/nethserver/${mid}"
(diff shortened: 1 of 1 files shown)


core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
```diff
@@ -38,6 +38,7 @@ if [[ -z $uid ]]; then # Rootfull module
     systemctl disable --now agent@"${mid}"
     if [[ ${preserve} == "no" ]]; then
         rm -rf "/var/lib/nethserver/${mid}"
+        podman volume rm $(podman volume ls | grep -o "${mid}-\w*")
```
DavidePrincipi, line 41: 1. Handle the case where grep produces an empty list
2. Ignore any error

```suggestion
         podman volume ls | grep -o "${mid}-\w*" | xargs -r -- podman volume rm || :
```

DavidePrincipi (approved):
:jack_o_lantern:

PR #309, opened 2022-11-03
Use Node v18 (enhancements)

andre8244:
Set `NODE_OPTIONS=--openssl-legacy-provider` in container as environment variable.

**About development inside VS Code container**
it looks like VS Code is ignoring the specified Node version inside `devcontainer.json` and always uses Node v16.
Since Node v16 doesn't accept `NODE_OPTIONS=--openssl-legacy-provider`, we need to set this environment variable inside VS Code integrated terminal before launching `yarn` commands. This has been configured inside `.vscode/settings.json`

### DIFF
--- a/core/ui/README.md
+++ b/core/ui/README.md
@@ -58,7 +58,6 @@ podman exec -ti ns8-core yarn storybook
 - Open Command Palette (`CTRL+SHIFT+P`) and type `Reopen in Container`
 - Open VS Code integrated terminal: `View > Terminal`
 - Change directory to `core/ui`
-- Enter `export NODE_OPTIONS=--openssl-legacy-provider` (this is needed for Node v18)
 - Enter one of the commands listed in [Develop NS8 UI on your workstation](#develop-ns8-ui-on-your-workstation), e.g. `yarn serve`
 
 Container configuration is contained inside `.devcontainer/devcontainer.json`.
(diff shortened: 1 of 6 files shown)


DavidePrincipi (approved):

core/ui/README.md
```diff
@@ -56,6 +56,9 @@ podman exec -ti ns8-core yarn storybook
 - Dev Containers uses Docker by default but you can configure it to use Podman: go to `File > Preferences > Settings`, search `dev containers docker path` and type `podman` as `Docker path`
 - Open `ns8-core` directory (the repository root) in VS Code, if you haven't already
 - Open Command Palette (`CTRL+SHIFT+P`) and type `Reopen in Container`
-- Open VS Code integrated terminal, change directory to `core/ui` and then type one of the commands listed in [Develop NS8 UI on your workstation](#develop-ns8-ui-on-your-workstation), e.g. `yarn serve`
+- Open VS Code integrated terminal: `View > Terminal`
+- Change directory to `core/ui`
+- Enter `export NODE_OPTIONS=--openssl-legacy-provider` (this is needed for Node v18)
```
Tbaile, line 61: To ease this, we can use a `.vscode/settings.json` file to allow the setting of an env variable *only* inside the terminal, this can be achieved with:
```
{
    "terminal.integrated.env.linux": {
        "NODE_OPTIONS": "--openssl-legacy-provider"
    }
}
```

andre8244: Good to know! Anyway we usually don't commit IDE-specific configurations. What do you think @DavidePrincipi ? Can we make an exception here?

DavidePrincipi: I thought we were making one already...

Tbaile (approved):

PR #312, opened 2022-11-21
Added default cluster name on core installation

Tbaile:
Edited the install.sh script to export in environment the random name of the container used to unpack the core.
Setting the name for the node just created before the cluster creation / node join.

### DIFF
--- a/core/install.sh
+++ b/core/install.sh
@@ -64,6 +64,8 @@ echo "Extracting core sources from ${core_url}:"
 mkdir -pv /var/lib/nethserver/node/state
 cid=$(podman create "${core_url}")
 podman export ${cid} | tar --totals -C / --no-overwrite-dir --no-same-owner -x -v -f - | LC_ALL=C sort | tee /var/lib/nethserver/node/state/coreimage.lst
+export CLUSTER_NAME
+CLUSTER_NAME=$(podman ps -a -f id="$cid" --format "{{.Names}}")
 podman rm -f ${cid}
 
 /var/lib/nethserver/node/install-core.sh "${modules[@]}"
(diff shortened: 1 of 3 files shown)


Tbaile:
Follow up question before anything else: I've noticed that multiple names are set throughout all the core. At the moment this PR updates only the `node/1/ui_name`, not the `cluster/ui_name`. We want to set them up both? I've got no clue on which value is used by 2FA nor how this name is used by 2FA.

DavidePrincipi:
The patch works for setting a _node_ default ui_name but I was expecting that only cluster/ui_name was set instead!

IIRC we're aiming to a _cluster_ default name!

https://trello.com/c/4eZRuYEL/233-cluster-name-label

Tbaile:
Oh well not a big deal, just changed where the name is set, should be fine now!

core/install.sh
```diff
@@ -64,6 +64,8 @@ echo "Extracting core sources from ${core_url}:"
 mkdir -pv /var/lib/nethserver/node/state
 cid=$(podman create "${core_url}")
 podman export ${cid} | tar --totals -C / --no-overwrite-dir --no-same-owner -x -v -f - | LC_ALL=C sort | tee /var/lib/nethserver/node/state/coreimage.lst
+export NODE_NAME
+NODE_NAME=$(podman ps -a -f id="$cid" --format "{{.Names}}")
```
DavidePrincipi, line 68: Let's call it CLUSTER_NAME!

```suggestion
export CLUSTER_NAME
CLUSTER_NAME=$(podman ps -a -f id="$cid" --format "{{.Names}}")
```

DavidePrincipi (approved):

edospadoni (approved):

PR #316, opened 2022-11-29
Fix userdel home locking

DavidePrincipi:
The command holds the lock on /etc/passwd and prevents other processes to modify it. Removing a big home directory holds the lock for a long time, blocking the removal of other modules.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
+++ b/core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
@@ -48,6 +48,9 @@ else # Rootless module
     if [[ ${preserve} == "yes" ]]; then
         userdel "${mid}"
     else
+        # Avoid /etc/passwd locking by erasing home directory separately:
+        # it is a long operation!
+        eval rm -rf ~"${mid}"
         userdel -r "${mid}"
     fi
 fi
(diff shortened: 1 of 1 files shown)


core/imageroot/var/lib/nethserver/node/actions/remove-module/50update
```diff
@@ -48,6 +48,9 @@ else # Rootless module
     if [[ ${preserve} == "yes" ]]; then
         userdel "${mid}"
     else
+        # Avoid /etc/passwd locking by erasing home directory separately:
+        # it is a long operation!
+        eval rm -rf ~"${mid}"
```
Amygos, line 53: Why use `eval`?

DavidePrincipi: Without it does not work. See tilde expansion: https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html

gsanchietti: That's eval is a little bit scary, maybe it's better to use `getent passwd <user>"?

gsanchietti (approved):

gsanchietti (approved):

Amygos (approved):

PR #318, opened 2022-12-05
PhoneHome Client Implementation

Tbaile:
Implemented phone home installation with the following approach:
 - the phone home should be started on leader node by a timer
 - dispatch 'get-facts' action to every node, isolating responsibility
 - every node sends back the information, gets merged by main process, then sent to server


Tbaile:
TODOs:
 - [ ] Remove task.run call, push directly task to redis task list
 - [ ] Implement base get-facts (even needed? best to check if in task-list of module there's a get-facts)

core/imageroot/etc/systemd/system/phonehome.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=PhoneHome
+
+[Service]
+Type=oneshot
+ExecStart=runagent /usr/local/bin/phonehome
+
+[Install]
+WantedBy=default.target
```
DavidePrincipi, line 9: This unit is activated by the timer: do not install it.

```suggestion
```

core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
```diff
@@ -0,0 +1,29 @@
...
+        key, value = line.rstrip().split('=')
+        platform[key] = value.strip('"')
+
+facts = {
+    'distro': {
+        'name': platform['ID'],
+        'version': platform['VERSION_ID']
+    },
+    # TODO: Replace when version is included in release
+    'version': '8.0.0'
```
DavidePrincipi, line 23: What about reading the core image version instead?

    /etc/nethserver/core.env

It can be a tag like 1.2.3, or a development branch name

docs/core/phone_home.md
```diff
@@ -0,0 +1,18 @@
...
+title: Phone Home
+nav_order: 15
+parent: Core
+---
+
+# Phone Home
+
+We're committed to being more aware of our products, how they are used and how they behave in different conditions and hardware.
+
+To do so, we actually need to expand how the Phone Home worked in NS6-NS7 by adding additional telemetry to be sent and processed for analysis.
```
DavidePrincipi, line 12: Could you add a command (e.g. `api-cli run get-facts`) to see a preview of what data is sent?

Can we state that data is always collected in anonymous form?

core/imageroot/etc/systemd/system/phonehome.timer
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=PhoneHome Trigger
+
+[Timer]
+OnCalendar=daily
```
DavidePrincipi, line 5: IIRC Systemd allows to add a random offset. Could you use it to avoid all clients send their payload at the same time?

core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
Tbaile: Sure, however, by using `core_url='ghcr.io/nethserver/core:ns8-stable` in the [install script](https://github.com/NethServer/ns8-core/blob/9a75d2008880a0735466aa489be14d7c17fb6453/core/install.sh#L28) this information is lost. Shall we update the install.sh script every time? Otherwise we can update the image with a label and then during the installation save it elsewhere.

docs/core/phone_home.md
Tbaile: > Could you add a command (e.g. api-cli run get-facts) to see a preview of what data is sent?

Was thinking the same thing once the PR went live :smile:, will add the line!

> Can we state that data is always collected in anonymous form?

As for now, yes, no info is actually fetched aside from `/etc/os-release`. We can always delete that statement.

core/imageroot/etc/systemd/system/phonehome.timer
Tbaile: Was not aware of the random offset, will do.

core/imageroot/var/lib/nethserver/cluster/actions/get-facts/50get
```diff
@@ -0,0 +1,57 @@
...
+}
+
+facts['uuid'] = redis_client.get('cluster/uuid')
+
+for key in redis_client.scan_iter('node/*/environment'):
+    # TODO: node_id not representative, change with NODE_UUID when available.
+    node_id = redis_client.hget(key, 'NODE_ID')
+
+    result = agent.tasks.run(
+        agent_id=f'node/{node_id}',
```
DavidePrincipi, line 40:
```suggestion
for key in redis_client.scan_iter('node/*/vpn'):
    agent_id = key.removesuffix('/vpn')

    result = agent.tasks.run(
        agent_id=agent_id,
```

core/imageroot/var/lib/nethserver/cluster/actions/get-facts/50get
```diff
@@ -0,0 +1,57 @@
...
+        action='get-facts',
+        extra={
+            'isNotificationHidden': True,
+        },
+        endpoint="redis://cluster-leader"
+    )
+
+    # returns error if task failed
+    agent.assert_exp(result['exit_code'] == 0, result['error'])
```
DavidePrincipi, line 49: Are you sure to abort if just one node fails? Shall we continue with the next one?

Tbaile: > Are you sure to abort if just one node fails?

In order to keep track of what happens to the cluster (more modules installed, node added, etc...) if a node fails, the phone-home will return that a node was removed. Since it's a thing that with the new installations appeared that we wanted to do, I've set a fail if the node fails to fetch info.

> Consider also to run all tasks in parallel with agent.runp_brief()

Already saw that, however only the error codes are returned with that function, but I need the whole output.

DavidePrincipi: If output is needed, you can still run in parallel with `agent.runp()`

The node task can fail because the node has been switched off. IIRC agent.run raises an exception in that case.

Tbaile: > If output is needed, you can still run in parallel with agent.tasks.runp()

Missed the function 10 rows above :sweat_smile:

> The node task can fail because the node has been switched off. IIRC agent.run raises an exception in that case.

So anything that happens to agent run, if failed, throws an exception? In the phonehome bin script, if the get-facts fails no exception were thrown during my test. Maybe it depends on what happens inside the task? :man_shrugging:

DavidePrincipi: If the action fails an exit code is returned. If the task fails an exception is thrown.
Action failure is caused by a step. Task failure is caused by communication errors.

core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
```diff
@@ -0,0 +1,37 @@
...
+with open('/etc/nethserver/core.env', encoding='utf-8') as file:
+    for line in file:
+        if 'CORE_IMAGE' in line:
+            print(core_version)
+            core_version = line[line.find(':') + 1:].strip()
+            break
+
+json.dump(
+    {
+        os.environ['AGENT_ID'].strip('node/'): {
```
Tbaile, line 28: Little note on this, I'm assuming that AGENT_ID will change when the identifier of the node is NODE_UUID, is that the case?

DavidePrincipi: I don't want to change how the AGENT_ID is built. What I don't grasp is how to obtain NODE_UUID. Use /etc/machine-id? Generate a random UUID? ...time will tell.

BTW [.strip()](/usr/libexec/nut-driver-enumerator.sh --get-device-for-service %i) is now what you expect. You might want .removeprefix() instead.

Tbaile: Well, come to think of it, we can access the `AGENT_ID/environment` on redis and read the NODE_ID from there. We can switch to UUID when the time comes. (assuming the UUID will be written there)

core/imageroot/etc/systemd/system/phonehome.timer
```diff
@@ -0,0 +1,11 @@
...
+Persistent=true
+AccuracySec=1us
+RandomizedDelaySec=86400
```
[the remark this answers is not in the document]
Tbaile: Pushing the timer on all 24 hours might have been too much :smile: 
The `AccuracySec=1us` is actually suggested by documentation itself when using big delays, I've got no issue to remove it.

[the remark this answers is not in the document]
DavidePrincipi: Yes they say "In order to _optimally_ stretch..."

However about AccuracySec they also say,

> To optimize power consumption, make sure to set this value as high as possible and as low as necessary.

So I remain of the idea that we do not need an accurate wake-up time: the default 1 minute period is good for our purpose.

DavidePrincipi (approved):

Tbaile:
Testing:
  - Install fresh NS8 installation using `phonehome-client` tag
  - Init cluster
  - Check if output of `api-cli run get-facts` validates with [server .json](https://raw.githubusercontent.com/NethServer/phonehome-server/0.3.1/resources/validation/facts/2022-12.json) (only `facts.node.1.version` will be incorrect)

DavidePrincipi:
> Check if output of `api-cli run get-facts` validates with [server .json](https://raw.githubusercontent.com/NethServer/phonehome-server/0.3.1/resources/validation/facts/2022-12.json) (only `facts.node.1.version` will be incorrect)

It is possible to add `validate-output.json` to `get-facts`: what do you think?

Tbaile:
> It is possible to add validate-output.json to get-facts: what do you think?

We could, it was there in the beginning, however
 - the output is not needed to anything other than the upstream server, which does the validation based on the $schema given
 - for branches installations, it will _always fail_ due to version not following semantic
 - there will be a duplication of code across projects, which It's pretty avoidable 
 
 Aside from this, we could  still copy the validation file, It's just a matter of a curl in the right place :smile:

gsanchietti (approved):

DavidePrincipi:
Let's merge it as-is

Amygos (approved):

PR #319, opened 2022-12-08
Agent function http_route_in_use and tcp_port_in_use

stephdl:
http_route_in_use : Test if an url exists with the traefik api, return true is used or false

tcp_port_exists: Test if the port is used or not, return true is used or false

```
import agent
domain = agent.http_route_in_use(domain='rocky9-2.local', path='', expectation=404)
webpath = agent.http_route_in_use(domain='127.0.0.1', path='/foo', expectation=404)
port=agent.tcp_port_in_use(5555)
```

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
+++ b/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
@@ -479,45 +479,20 @@ def get_smarthost_settings(rdb):
         }
     return data
 
-def http_route_in_use(domain='127.0.0.1', path='', expectation=404):
-    """ we search if the domain (foo.com) or the webpath (/foo) is used or not by a service
-        return true if the route is used, false is the route is not used"""
-
-    # We cannot use the domain of the server else 
-    # we cannot reach the cluster-admin
-    if domain == socket.getfqdn():
-        return True
-
-    http = 0
-    https = 0
-
-    # do not verify the certificate
-    ctx = ssl.create_default_context()
-    ctx.check_hostname = False
-    ctx.verify_mode = ssl.CERT_NONE
-    
-    # try on http
+def http_route_in_use(domain=None, path=''):
+    """we search if the domain (foo.com) or the webpath (/foo) is used or not by a service
+    return true if the route is used, false is the route is not used"""
     try:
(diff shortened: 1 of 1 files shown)


core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -474,3 +479,60 @@ def get_smarthost_settings(rdb):
...
+        urllib.request.urlopen(req, context=ctx)
+    except urllib.error.HTTPError as d:
+        https = d.code
+
+    if http == expectation and  https == expectation:
+        # there is no website on http or https -> OK
+        return False
+    else:
+        # path exists -> nok
+        return True
```
DavidePrincipi, line 520: That's all...

```suggestion
def http_route_in_use(domain=None, path='', expectation=404):
    """ we search if the domain (foo.com) or the webpath (/foo) is used or not by a service
        return true if the route is used, false is the route is not used"""
    try:
        req = urllib.request.Request('http://127.0.0.1'+path)
        if domain:
            req.add_header('Host', domain)
        urllib.request.urlopen(req, context=ctx)
    except urllib.error.HTTPError as e:
        if e.code == expectation:
            # there is no website on http or https -> OK
            return False

    return True
```

DavidePrincipi: ...I guess the urlopen() function could also succeed without raising any exception. What happens with status code 200?

stephdl: yes only 404 is the return that the route does not exist

DavidePrincipi: Edit: added `method="HEAD"` to avoid fetching page contents

DavidePrincipi (approved):

core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -32,6 +32,11 @@
 import shlex
 import warnings
 import agent.tasks
+import socket
+import errno
+import urllib.request
+import urllib.error
+import ssl
```
DavidePrincipi, line 39:
```suggestion
```

PR #321, opened 2022-12-13
Fix module missing permissions

DavidePrincipi:
Short bug description: if Webtop is installed before Mail, the role `mailadm` is not granted. See [Trello card](https://trello.com/c/CMUA45Zc/239-missing-authorizations).

This PR defines two new Redis keys:

- `cluster/authorizations/{agent_id}` SET, to store the module `org.nethserver.authorizations` label 
- `cluster/module_node` HASH, to store the node_id where a module is installed. This is safer than relying on the environment NODE_ID value

Stored data is used to re-calculate a module permissions when a new module is added to the cluster.
Authorizations are never updated/upgraded. Only those given at install time are considered.


core/imageroot/usr/local/agent/pypkg/cluster/grants.py
```diff
@@ -99,3 +99,37 @@ def alter_user(rdb, user, revoke, role, on_clause):
...
+        mnode_id = rdb.hget('cluster/module_node', kmid)
+        add_module_permissions(rdb, kmid, authorizations, mnode_id)
+
+def add_module_permissions(rdb, module_id, authorizations, node_id):
+    """Parse authorizations and grant permissions to module_id"""
+    for authz in authorizations:
+        if '*' in authz:
+            print(agent.SD_WARNING + f'Cannot grant authorization "{authz}" to {module_id}. Wildcard is not allowed, use @any suffix instead.')
+            continue
+
```
gsanchietti, line 116: If wildcard is not allowed, why even bother?
```suggestion
```

core/imageroot/usr/local/agent/pypkg/cluster/grants.py
```diff
@@ -99,3 +99,37 @@ def alter_user(rdb, user, revoke, role, on_clause):
...
+        xagent, xrole = authz.split(':') # e.g. traefik@node:routeadm
+
+        if xrole == 'owner':
+            print(agent.SD_WARNING + f'Cannot grant authorization "{authz}" to {module_id}. Role "{xrole}" is forbidden.')
+            continue
```
gsanchietti, line 121: What are valid values for `xrole`?

gsanchietti (commented):
Also the description states:

> Authorizations are never updated/upgraded.

Why not executing this function also on upgrade? Of course, the new authorization labels should be saved first.

core/imageroot/usr/local/agent/pypkg/cluster/grants.py
DavidePrincipi: I don't understand the code suggestion. The `for` statement can't be removed!

The whole function body is a cut-paste from the previous add-module action implementation.

I'm removing the `print` statement anyway because it is not good in a library function.

gsanchietti: I selected the wrong line.
I'm proposing to delete lines from 113 to 115

core/imageroot/usr/local/agent/pypkg/cluster/grants.py
gsanchietti: Ok, so I do not understand why `owner` is not allowed.

core/imageroot/usr/local/agent/pypkg/cluster/grants.py
```diff
@@ -99,3 +99,36 @@ def alter_user(rdb, user, revoke, role, on_clause):
...
+        kmid = kauth_module.removeprefix('cluster/authorizations/module/')
+        authorizations = rdb.smembers(kauth_module) or []
+        mnode_id = rdb.hget('cluster/module_node', kmid)
+        add_module_permissions(rdb, kmid, authorizations, mnode_id)
+
+def add_module_permissions(rdb, module_id, authorizations, node_id):
+    """Parse authorizations and grant permissions to module_id"""
+    for authz in authorizations:
+        if '*' in authz or '[' in authz or '?' in authz:
+            continue # Redis glob patterns are forbidden
```
gsanchietti, line 114: Now it's better.
But since there is no strong validation for `authz` why not removing the check?

DavidePrincipi: I want to keep the check and implement a simple validation rule.

gsanchietti: :shrug:

core/imageroot/usr/local/agent/pypkg/cluster/grants.py
DavidePrincipi: I moved those lines to a point where the validation rule can be enforced

See commit https://github.com/NethServer/ns8-core/pull/321/commits/4dcb19936290277aadf1c2f056091f2e4b719190

DavidePrincipi:
> Why not executing this function also on upgrade? Of course, the new authorization labels should be saved first.

Yes it makes sense. I preferred to focus on fixing the bug here. Let's discuss it separately.

gsanchietti (approved):

PR #322, opened 2022-12-13
Allow addition of pip packages

DavidePrincipi:
This PR allows rootless modules to add more Python packages to the base  `pyenv` agent environment: they are installed under the module home directory, with `pip install --user ...` 

References:

https://peps.python.org/pep-0405/#isolation-from-system-site-packages

> [PEP 370](https://peps.python.org/pep-0370) user-level site-packages are considered part of the system site-packages for venv purposes: they are not available from an isolated venv, but are available from an include-system-site-packages = true venv.

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/install-core.sh
+++ b/core/imageroot/var/lib/nethserver/node/install-core.sh
@@ -49,7 +49,7 @@ fi
 
 echo "Setup Python virtual environment for agents:"
 core_dir=/usr/local/agent/pyenv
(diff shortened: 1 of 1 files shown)


stephdl (approved):

Amygos:
Can you provide also an example of how a module can install an additional package?

Amygos:
Can you provide also an example of how a module can install an additional package?

Amygos (approved):

gsanchietti (approved):

PR #325, opened 2022-12-19
add-module. Define the `selfadm` role implicitly

DavidePrincipi:
Every module is authorized to run configure-module on itself. If this is undesirable, or if more actions are needed it is possible to modify the Redis SET key that defines the `selfadm` role:

    module/{module_id}/roles/selfadm
    
See also the attached documentation.

### DIFF
--- a/core/imageroot/usr/local/agent/actions/create-module/10selfadm_role
+++ b/core/imageroot/usr/local/agent/actions/create-module/10selfadm_role
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+#
+# Copyright (C) 2022 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+exec 1>&2
+set -e
+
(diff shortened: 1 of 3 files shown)


Amygos:
What is the rationale for limiting the default allowed actions to `configure-module`? In my opinion, a module can call all the actions defined by himself to himself.

DavidePrincipi:
Of course we can also grant everything with a "*", but before making that step I'd think a while.

As this is a matter of granting permissions I'm trying to give as little of them as possible, to avoid abuses of self-task calls. Such calls are a complex thing that may fail and thus do no suit every situation.

The configure-module, if it exists is the action that completes the module initialization after create-module and is likely to be invoked in restore-module and clone-module, which are similar but different ways to construct a module instance. This is our experience so far, thus I'm just limiting to it.

Amygos:
Ok. I understand your concerns about the possible unknown side effects of this type of operation. For now, we can start with this approach.

Amygos (approved):

PR #327, opened 2022-12-28
Fix clone-module auth error

DavidePrincipi:
The file apitoken.cache must be excluded from transfer-state (if it exists). Otherwise the destination module will use the source module credentials to authenticate itself.


DavidePrincipi:
CentOS stream is still failing for the empty log test :sob:

PR #328, opened 2023-01-09
Updated yarn to v2

Tbaile:
Due to node 18 migration and [new yarn installation process](https://yarnpkg.com/getting-started/install), if you happen to fresh install the node\_modules you'll likely won't be able to. Yarn might automatically migrating the whole lock file to v2. This PR handles the migration process.


andre8244 (approved):
LGTM: working on both Podman and VS Code Dev Containers

PR #331, opened 2023-01-13
Fix phonehome test case

DavidePrincipi:
Error message

    nethserver != 'nethserver'

See also

https://github.com/NethServer/ns8-core/actions/runs/3910845890/jobs/6683529749#step:11:103

### DIFF
--- a/core/tests/10__cluster_sanity/31__phonehome.robot
+++ b/core/tests/10__cluster_sanity/31__phonehome.robot
@@ -11,5 +11,5 @@ Check correct output from get-facts
     ${response} =    Run task    cluster/get-facts    null
     Should Not Be Empty    ${response['$schema']}
     Should Not Be Empty    ${response['uuid']}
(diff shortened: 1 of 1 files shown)


DavidePrincipi:
@Tbaile red lights :sob:

Tbaile:
Seems that we have more than one failure.
While the fix is correct comparing strings, tests skips entirely NS8 installation due to [this line](https://github.com/NethServer/ns8-core/blob/07b13f1f051c9a172a01521d81bdc0594ddb7cc8/.github/workflows/tests.yml#L49) that directly [reflects on this workflow](https://github.com/NethServer/ns8-github-actions/blob/7580d6b469065e6879d16315d06ea9c129f1b45b/.github/workflows/test-on-digitalocean-infra.yml#L134).
The test are ultimately right, cannot ask a get-fact on a uninitialized node, due to UUID missing and service/timer not enabled.

PR #333, opened 2023-01-26
Custom firewall zones

gsanchietti:
Add and remove custom firewall zones.

Required by https://github.com/NethServer/ns8-nextsec-controller/pull/3

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
+++ b/core/imageroot/usr/local/agent/pypkg/agent/__init__.py
@@ -422,6 +422,21 @@ def add_public_service(name, ports):
     )
     return response['exit_code'] == 0
 
+def add_custom_zone(name, interface, ports=[], rules=[]):
+    node_id = os.environ['NODE_ID']
+    response = agent.tasks.run(
+        agent_id=f'node/{node_id}',
+        action='add-custom-zone',
+        data={
+            'zone': name,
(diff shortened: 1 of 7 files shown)


gsanchietti:
> Can `add-custom-zone` run successfully even if `interface` is not present?

Not with current implementation.

> Does firewalld enforce the zone definition even if `interface` is dynamic, in other words it is not present at boot and comes up at a later time?

This is what I expect from any decent firewall engine but I never tested firewalld. I will leave the check to @Tbaile

Tbaile:
> This is what I expect from any decent firewall engine but I never tested firewalld. I will leave the check to @Tbaile

Firewall correctly allows zone setting without interface attached, tested on rocky `9.1` and firewalld `1.1.1-3.el9`

Tbaile (approved):

Tbaile:
Cannot test properly due to permission issues with `add-custom-zone`, it seems that the agent forbids the request even if [authorization is granted to node](https://github.com/NethServer/ns8-core/pull/333/files#diff-a372f7379c51327ddea6d57692754ec4410489ce9408f05514db24011b855b65R181)

gsanchietti:
> > without interface attached
> 
> What happens when the interface comes up? Does it enforce the interface rules?

Rules are correctly enforced even after the reboot when the VPN is not stil up&running.

PR #334, opened 2023-02-08
Fix Static Build

Tbaile:
Due to issues with static building in `alpine`, I've tried to fix the problem due to the `musl` linker missing from the final executable.

These two options were considered, but they utterly were failures:
- :x: trying to add the official alpine package `build-base` still makes `go build` use `musl` as linker, so no success was reached.
- :x: dynamically building the executable is faster but introduces all sorts of the dependency issues, since the binary is not used inside a container.

A workaround for the soulution is to static build the executable inside the `bullseye` flavor of the `golang` container.

During building/running, no warning is provided due to missing libraries, `ldd` output seems fine.

### DIFF
--- a/core/build-image.sh
+++ b/core/build-image.sh
@@ -11,8 +11,7 @@ if ! buildah containers --format "{{.ContainerName}}" | grep -q gobuilder-core; 
     echo "Pulling Golang runtime..."
     golang_cache_path="${PWD}/.golang-cache"
     mkdir -vp "${golang_cache_path}/{mcache,bcache}"
-    buildah from --name gobuilder-tmp docker.io/library/golang:1.18-alpine
-    buildah run gobuilder-tmp apk add g++ gcc
+    buildah from --name gobuilder-tmp docker.io/library/golang:1.19.5-bullseye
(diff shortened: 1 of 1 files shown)


gsanchietti (changes requested):
The api can't start anymore on a rocky linux distro:
```
Feb  9 09:11:18 rl1 kernel: api-server[59400]: segfault at 8 ip 00000000004074e0 sp 00007ffc3f2f07b0 error 4 in api-server[407000+76f000]
Feb  9 09:11:18 rl1 kernel: Code: 68 48 00 00 00 e9 60 fb ff ff ff 25 02 3b e1 00 66 90 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 <48> 8b 04 25 08 00 00 00 0f 0b 48 8b 04 25 10 00 00 00 0f 0b 41 bc
```

Tbaile:
@gsanchietti could you try again? Seems on our side that the installation works, however before merging we found out that alpine build appears to be working with the current build params, so we'll try again with alpine later

gsanchietti (approved):

PR #342, opened 2023-02-22
Agent set-env unset-env commands deprecation

DavidePrincipi:
This PR deprecates the client use of agent `set-env` and `unset-env` commands.

The Python `agent` library do not use those command any more. Instead read/writes the `./environment` file on its own.

Existing Bash scripts relying on the deprecated commands need to be reimplemented:

- [set-env](https://github.com/search?q=set-env+org%3ANethServer&type=code)
- [unset-env](https://github.com/search?q=unset-env+org%3ANethServer&type=code)

After the above code is fixed, the commands can be removed from the agent.

### DIFF
--- a/core/imageroot/usr/local/agent/actions/clone-module/90finalize
+++ b/core/imageroot/usr/local/agent/actions/clone-module/90finalize
@@ -8,18 +8,15 @@
 set -e
 exec 1>&2 # Redirect any output to the journal (stderr)
 
+# Clean up clone-module state files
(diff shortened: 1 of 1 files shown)


Amygos:
What are the motivations behind this OR? What is wrong with the current implementation and what problems wants to solve this PR?

Amygos (approved):

gsanchietti (approved):

PR #346, opened 2023-02-27
Hidden users and groups definition

DavidePrincipi:
Define two sets of account provider entries
- the set of hidden users
- the set of hidden groups

Elements of those sets must be configured as not visible and not accessible to LDAP clients like the user-interface and additional modules. They must be like they do not exist at all.

Applications can obtain prepared LDAP search filter clauses from the `agent.ldapproxy` Python module. The sets are added to the return value of the `.get_domain()` method.

~~The actual set elements are stored in the Redis DB under the service (`/srv`) keys published by each account provider instance *and* under the following domain-wide Redis keys:~~

- ~~`cluster/user_domain/ldap/{domain}/hidden_users`~~
- ~~`cluster/user_domain/ldap/{domain}/hidden_groups`~~

~~The latter are applied also to external account providers and are for local cluster customizations (an UI to change them is *not* provided though).~~

<add>The sets are hardcoded in the Python `agent.ldapproxy` module</add>

----
https://trello.com/c/a2a9oYn2/210-core-p1-hide-builtin-users-and-groups

- https://github.com/NethServer/ns8-mail/pull/35
- https://github.com/NethServer/ns8-core/pull/346
- https://github.com/NethServer/ns8-ldapproxy/pull/6
- https://github.com/NethServer/ns8-samba/pull/7
- https://github.com/NethServer/ns8-openldap/pull/8

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/agent/ldapproxy.py
+++ b/core/imageroot/usr/local/agent/pypkg/agent/ldapproxy.py
@@ -54,8 +54,28 @@ class Ldapproxy:
                 conf[key] = self.domains[domain][key]
 
             conf['port'] = self.domains[domain]['listen_port']
-            conf['hidden_users'] = self.domains[domain]['hidden_users']
-            conf['hidden_groups'] = self.domains[domain]['hidden_groups']
+            if conf['schema'] == 'ad':
+                conf['hidden_users'] = [
+                    'Guest',
+                    'krbtgt',
+                    'ldapservice',
+                ]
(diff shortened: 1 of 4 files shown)


gsanchietti (approved):
A little bit complex.
If not too hard, I'd remove the customization options:
- `cluster/user_domain/ldap/dom.test/hidden_users`
- `cluster/user_domain/ldap/dom.test/hidden_groups`

gsanchietti (approved):

stephdl (approved):

PR #348, opened 2023-03-01
Fix unset/set-env with absolute environment path

DavidePrincipi:
Fixes regression from PR #342. If the step working directory is not under AGENT_STATE_DIR the ./environment file is not found and agent.set_env() aborts. This happens within the core implementation of the update-module action.

This commit open the environment file with an absolute path, so it can be invoked from a different working directory.


gsanchietti (approved):
The code looks code

PR #353, opened 2023-03-09
Once installed, the nethserver image could be removed

stephdl:
With add-module we install  nethserver image of container and these images are not removed from the local repository for the root user.

This PR attempt once installed to remove them


gsanchietti:
I agree with @DavidePrincipi .

@stephdl how much space is freed on a clean NS8 installation?

Amygos:
It was not clear to me that you meant the container of the module definition, in this case, I either don't see any problem with leaving the container images on the disk after the module installation. As mentioned before they usually are tiny and can act as a cache or metadata storage. Also the IIRC the `remove-module`  action remove it on module uninstall.

Why do you think this PR is necessary? What is the problem that you have encountered?

stephdl:
> Why do you think this PR is necessary? What is the problem that you have encountered?

https://nethserver.github.io/ns8-core/quickstart/

image


maybe it could help this, when you try to install an image with the same tag, it does not download it again, it uses the local image, if it is wrong you reinstall the same issue

Amygos:
Ok, I can understand the frustration, but as it is written in the developer documentation:

> If the given image is already present in the local Podman storage, no remote download occurs and the local image is used instead. During development this might be unwanted and to work around this behavior execute the following command in every cluster node, before add-module

In a production environment it is expected behavior, the same tag equals the same software. Maybe we can add an action at the cluster level that purge a specific tag in all nodes of the cluster? To use for maintenance or during development.

stephdl:
Sure I understand that `dev !== noob` and it was only the following of the previous PR I did to remove the application container that it weight more.

However if I install ns8-roundcubemail at 0.0.1 version and I upgrade to 0.0.2, then we remove by the agent script the roundcubemail container but the nethserver image will stay on the server for no purpose.

Nonetheless it is not a big amount of developer time, so I propose to close the PR :D

PR #357, opened 2023-03-13
Fix rsync client race condition

DavidePrincipi:
The TERM signal must be raised after the rsync reply is sent back otherwise the client might raise an error, depending on random factors.


gsanchietti:
The migration works, but are you sure this change doesn't break anything else?
I'd close this PR for now.

gsanchietti:
As you wish.

PR #364, opened 2023-03-22
api: fix acquire-fsmo in remove-internal-provider

gsanchietti:
acquire-fsmo requires admin credentials

Samba PR: https://github.com/NethServer/ns8-samba/pull/12

See also https://trello.com/c/f6u0mII7/327-core-p1-ui-for-samba-domain-removal


gsanchietti:
> ..and we already have them top

But it's not documented that credentials are required! :(

PR #372, opened 2023-03-31
install.sh. Test if FQDN is resolvable

DavidePrincipi:
The /etc/hosts record existence check might fail but the address is
still resolvable: for instance, RHEL-derived distros resolve the
hostname locally with the NSS "myhostname" module.


gsanchietti (changes requested):
The problem is not if the host is resolvable. The problem is the resolutions is low.

PR #382, opened 2023-04-14
Suppress repeated Agent BRPOP errors

DavidePrincipi:
The main agent task loop watches the task queue and produces repeated error log lines (one every 5 seconds) if credentials are wrong or connection is down. This commit adds a comparison of the error log message with the previous loop run and prints the message only if they differ.

For instance, in a multi-node cluster the `cluster` agent of worker nodes typically produces lines like this one:

    Task queue pop error: READONLY You can't write against a read only replica.

They can be silenced with this PR.

Another example: if the leader is unreachable every agent starts complaining, flooding the log with repeated entries.

Refs https://trello.com/c/9EG8GJia/326-core-p2-multi-node-ui

----

Fixes also the WRONGPASS error

Refs https://trello.com/c/8tjQ5S7Y/90-core-p3-wrongpass-invalid-username-password-pair

### DIFF
--- a/core/agent/htask.go
+++ b/core/agent/htask.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"encoding/csv"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


core/agent/htask.go
```diff
@@ -339,20 +340,37 @@ func listenActionsAsync(brpopCtx context.Context, complete chan int) {
...
+			// poll timeout, it's ok: start a new cycle
+			lastPopErr = nil // Clear the last error buffer
 			continue
 		} else if brpopCtx.Err() != nil {
 			break
 		} else if popErr != nil {
-			log.Print(SD_ERR+"Task queue pop error: ", popErr)
+			// Avoid error log repetitions: print the error only if it is
+			// different from the value in the last error buffer
+			if lastPopErr == nil || popErr.Error() != lastPopErr.Error() {
```
gsanchietti, line 364: Why are you checking if `lastPopErr` is null? I do not know all the code, but it seems to me is not useful: if the current error is different from the last one, just print it
```suggestion
			if popErr.Error() != lastPopErr.Error() {
```

gsanchietti (approved):

PR #389, opened 2023-04-19
Avoid 2FA lockout on leader promotion

DavidePrincipi:
As the new leader node has no access to 2FA secrets stored in the old node filesystem, we must disable 2FA to avoid admin lockout.


edospadoni (approved):

gsanchietti (approved):

gsanchietti:
Remember to add a note inside the UI.

PR #391, opened 2023-04-20
Refactor Redis ACLs

DavidePrincipi:
- Prevent access to sensitive information stored in Redis keys matching `*/task/*`
- Store task information under an new key pattern space. From `*/task/*` to `task/*`
- Use node credentials for DB replication, instead of anonymous credentials (`default` user)

Note that the API server URL paths do not change with this PR. Only Redis key patterns are affected.

- [x] update database docs


core/build-image.sh
```diff
@@ -5,6 +5,7 @@ cleanup_list=()
 trap 'rm -rf "${cleanup_list[@]}"' EXIT
 images=()
 repobase="${REPOBASE:-ghcr.io/nethserver}"
+redis_tag=7.2-rc-alpine
```
gsanchietti, line 8: I do not like using the rc version.
From the changelog:
```
Redis Release Candidate (RC) versions are early versions that are made available
for early adopters in the community to test them. We do not consider
them suitable for production environments.
```

Can we switch to 7.0?

gsanchietti (changes requested):
Regarding the whole refactor: I do not know enough the code to express a judgment.
But we should not use a release candidate for our redis which is a core part of the cluster.

gsanchietti (approved):
:crossed_fingers:

edospadoni (approved):

PR #397, opened 2023-05-05
Fix API regression after permissions refactor

DavidePrincipi:
The Redis `user/` key prefix is no more accessible with unprivileged access.
Use cluster credentials to read it.

See
- #396 
- https://trello.com/c/K90JTp6S/354-redis-key-access-perms-regression


gsanchietti (approved):
The fix was way much simpler than I thought! :+1:

PR #420, opened 2023-07-05
Remove set-env/unset-env agent commands

DavidePrincipi:
* Remove the deprecated agent commands `set-env` `unset-env`
* Unconditionally copy the state/environment file contents  to  `AGENT_ID + /environment` Redis HASH key at the end of the action run

With this PR the agent behavior stops to modify the state/environment file. It only keeps reading it to initialize the step process environment, and it copies the environment file contents to Redis at the end of the action run.

See also
- https://github.com/NethServer/ns8-traefik/pull/32
- https://github.com/NethServer/ns8-loki/pull/4
- https://github.com/NethServer/ns8-webtop/pull/22

References https://trello.com/c/jBkc6Qwt/365-core-p1-agent-set-env-deprecation

### DIFF
--- a/docs/modules/agent.md
+++ b/docs/modules/agent.md
@@ -97,7 +97,7 @@ Python example:
 ```python
 import agent
 
-agent.set_env('MYVAR', 'value')
+agent.set_weight('05longstep', 2)
 ```
 
 ## Validation


gsanchietti (approved):
From a quick search it seems to module is still using it :crossed_fingers:

PR #421, opened 2023-07-06
Tag the latest stable release as "ns8-stable"

DavidePrincipi:
Introduce a git tag "ns8-stable", with the same semantics of "ns8-stable" image tag .

When a stable _semver_ tag is pushed, move the "ns8-stable" git tag to the tagged commit.

This tag is needed because installation docs point to the `main` branch: `install.sh` is actually missing a _testing_ stage.

References https://community.nethserver.org/t/installation-fails-on-debian-11/21905

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -1,6 +1,10 @@
 name: core
 on:
   push:
+    branches:
+      - '**'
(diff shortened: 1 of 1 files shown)


gsanchietti (approved):
Seems good

PR #424, opened 2023-07-25
Agent must ignore exec() errors

DavidePrincipi:
**Generic scenario**

The PR changes the behavior of agent action and event handlers to skip exec() errors and just write a warning to the log.

**Real case scenario**

During core updates the running action `update-core` might be modified and a step executable script suddenly disappears. In this case the action should not fail but just report a warning.

A similar situation might arise in module updates too.


gsanchietti (approved):
Much more complex than I thought, but it seems ok to me.

PR #425, opened 2023-07-27
Block Redis startup until replication is complete

DavidePrincipi:
During "join-node" ensure `apply-vpn-routes` runs after replica synchronization is complete.

See bug report here: https://trello.com/c/7GXUKgrn/414-core-p1-vpn-routes-not-refreshed-by-join-node


gsanchietti (commented):
The scripts acl-load and apply-vpn-routes are not required anymore?

gsanchietti (approved):

PR #431, opened 2023-08-07
Update install-finalize.sh

DavidePrincipi:
Clarify the use of FQDN to reach the node is just an example. We cannot set as installation pre-requirement that FQDN is resolvable by DNS.

https://github.com/NethServer/ns8-docs/pull/25

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/install-finalize.sh
+++ b/core/imageroot/var/lib/nethserver/node/install-finalize.sh
@@ -160,7 +160,7 @@ Finally, access the administration UI at:
 
    https://<hostname_or_IP>/cluster-admin/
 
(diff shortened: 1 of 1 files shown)


core/imageroot/var/lib/nethserver/node/install-finalize.sh
```diff
@@ -158,8 +158,14 @@ B. To initialize this node as a cluster leader run:
 
 Finally, access the administration UI at:
 
+   https://<hostname_or_IP>/cluster-admin/
+
+For instance, if $(hostname -f) is resolveable
```
gsanchietti, line 163:
```suggestion
For instance, if $(hostname -f) is resolvable 
```

stephdl (approved):

PR #458, opened 2023-09-07
build-disk-images: remove Debian

gsanchietti:
Retire the image until it's ready to be used on the cloud


gsanchietti:
:scream: wrong commit!

PR #470, opened 2023-10-13
Extend agent.ldapclient Python module

DavidePrincipi:
- Protect the LDAP search filter from special chars injection
- Return additional attributes as raw ldap3 object for login/change-password purposes


gsanchietti (commented):
:shrug:

PR #516, opened 2023-11-07
Run command under rootless module environment

DavidePrincipi:
The root user can run commands under the environment of any rootfull module running in the local node.

This PR makes possible to run commands also under the environment of local rootless modules. For example

    runagent -m traefik1 systemctl --user status traefik

With this new feature, it is no longer needed to use the SSH client to run commands of rootless modules.

The implementation changes UID/GID of the root process with `runuser`, and properly set XDG_RUNTIME_DIR for Systemd runtime setup.

### DIFF
--- a/docs/core/backup_restore.md
+++ b/docs/core/backup_restore.md
@@ -163,7 +163,7 @@ api-cli run run-backup --data '{"id":1}'
 ```
 
 For debugging purposes, you can also launch systemd units:
-- rootless container, eg. `dokuwiki1`: `ssh dokuwiki1@localhost "systemctl --user start backup1.service"`
+- rootless container, eg. `dokuwiki1`: `runagent -m dokuwiki1 systemctl --user start backup1.service`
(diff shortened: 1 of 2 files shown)


stephdl (approved):

core/imageroot/usr/local/bin/runagent
```diff
@@ -65,9 +66,30 @@ elif os.path.isdir(f"/var/lib/nethserver/{mid}"):
...
+    rargv = list(sys.argv)
+    for targ in ["-m", "--module-id"]:
+        try:
+            idx = rargv.index(targ)
+        except ValueError:
+            continue
+        rargv.pop(idx) # drop -m
+        rargv.pop(idx) # drop the value
+
+    xdg_runtime_dir=f"/run/user/{module_uid}"
```
gsanchietti, line 87: Does this directory exists as soon as user lingering is enabled?

gsanchietti: :+1:

gsanchietti (approved):

PR #518, opened 2023-11-09
Local backups in cluster via WebDAV

DavidePrincipi:
- Start a Systemd `restic-webdav.socket` unit on any cluster node. Each node becomes a backup repository, accessible from other cluster nodes from the Wireguard VPN, via WebDAV protocol. Restic clients uses `rclone` as remote driver to talk with the WebDAV service.
- Implement the WebDAV service with `rclone serve webdav`. Actually a patched build of `rclone` is generated to support the Systemd socket activation protocol. The patch has been submitted to Rclone: https://github.com/rclone/rclone/pull/7392
- Extend existing APIs with the new `cluster` backup backend type: `create-backup-repository`, `alter-backup-repository`
- New cluster action `list-cluster-backup-endpoints`, to get the list of WebDAV service URLs
- Required RC1 Release Notes to start the .socket unit in existing installations: https://github.com/NethServer/ns8-docs/pull/35

**Examples**

Use node 2 as repository

```
api-cli run add-backup-repository --data '{"name":"Node 2 test repository", "provider": "cluster", "password":"", "url":"webdav:http://10.5.4.2:4695", "parameters":{}}'
```

Check the WebDAV server is responsive, expect HTTP 200:

    curl -I http://cluster-localnode:4694

### DIFF
--- a/core/imageroot/var/lib/nethserver/cluster/actions/add-backup-repository/10validate
+++ b/core/imageroot/var/lib/nethserver/cluster/actions/add-backup-repository/10validate
@@ -51,7 +51,7 @@ renv = dict(os.environ) # copy the environment
 renv.update(request['parameters']) # merge parameters
 renv['url'] = request['url'] # add url
 renv['provider'] = request['provider'] # add provider
-pvalidate = subprocess.run(['rclone-wrapper', '0', 'size', '-q', '--json', 'REMOTE_PATH'], text=True, capture_output=True, env=renv)
+pvalidate = subprocess.run(['rclone-wrapper', '0', '--low-level-retries', '1', '--contimeout', '10s', 'size', '-q', '--json', 'REMOTE_PATH'], text=True, capture_output=True, env=renv)
 if pvalidate.returncode != 0:
     errors.append({'field':'parameters','parameter':'parameters','value':pvalidate.stderr,'error':'backup_repository_not_accessible'})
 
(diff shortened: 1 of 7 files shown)


core/build-image.sh
```diff
@@ -119,18 +119,45 @@ buildah commit "${container}" "${repobase}/${reponame}"
...
+mkdir -p /srv/build
+cd /srv/build
+rm -rf rclone-*
+curl -L https://github.com/rclone/rclone/archive/refs/tags/v1.62.2.tar.gz | tar xvz
+cd rclone-*
+echo "Apply patch for systemd-activation to $(pwd)..."
+curl -L https://github.com/rclone/rclone/commit/0f15efae616354cc887063835c8dfa510488f9a6.patch | patch -p1
+export GO111MODULE=on
+export CGO_ENABLED=0
+go build
```
gsanchietti, line 136: I think this will be hardly maintainable in the long run.
As an example, when we are going to bump the rclone version, the path could be invalid for the new release.
I also do not see any real advantage on having socket activation: you can just spare few megabytes of RAM when the backup is not running.

core/imageroot/etc/systemd/system/restic-webdav.service
```diff
@@ -0,0 +1,29 @@
+[Unit]
+Description=Restic WebDAV (rclone)
```
gsanchietti, line 2: There is a bit of confusion with the naming here: is it rclone or restic server?

I'd rename it Rclone WebDAV server.

gsanchietti (changes requested):

gsanchietti (approved):
Thanks for the changes. We need to remember to document how to map the volume to an extra disk.

stephdl (approved):

PR #521, opened 2023-11-13
Backup Handle error focus for properties

stephdl:
We need to specify differently how to retrieve the correct property to fix the focus when a field is not validated correctly by the Json schema

image

image

### DIFF
--- a/core/ui/public/i18n/en/translation.json
+++ b/core/ui/public/i18n/en/translation.json
@@ -544,10 +544,10 @@
     "samba_placeholder":"Share name",
     "smb_short": "SMB",
     "samba_tooltips":"Only shares with SMB2/3 protocols are supported. Deprecated SMB1/CIFS shares are not secure and cannot be accessed.",
-    "cluster": "Local storage",
-    "cluster_short": "Cluster",
+    "cluster_short": "Local storage",
     "node": "Node",
-    "cluster_placeholder": "Choose a node in the cluster"
+    "cluster_placeholder": "Choose a node in the cluster",
+    "parameters_format": "Incorrect value"
   },
   "calendar": {
     "days": "Days",
--- a/core/ui/src/components/backup/AddRepositoryModal.vue
+++ b/core/ui/src/components/backup/AddRepositoryModal.vue
@@ -385,6 +385,7 @@
 <script>
 import { UtilService, TaskService, IconService } from "@nethserver/ns8-ui-lib";
 import to from "await-to-js";
+import last from "lodash/last";
 
 export default {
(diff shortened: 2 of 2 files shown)


core/ui/public/i18n/en/translation.json
```diff
@@ -544,10 +544,10 @@
     "samba_placeholder":"Share name",
     "smb_short": "SMB",
     "samba_tooltips":"Only shares with SMB2/3 protocols are supported. Deprecated SMB1/CIFS shares are not secure and cannot be accessed.",
-    "cluster": "Local storage",
```
andre8244, line 547: This i18n string is used in the first step of the wizard (repo type selection), let's not remove it

core/ui/public/i18n/en/translation.json
```diff
@@ -544,10 +544,10 @@
...
-    "cluster_short": "Cluster",
+    "cluster_short": "Local storage",
     "node": "Node",
-    "cluster_placeholder": "Choose a node in the cluster"
+    "cluster_placeholder": "Choose a node in the cluster",
```
andre8244, line 549: The name of this i18n key is a bit too generic, let's rename it to something like `"choose_cluster_node"`, or let's remove it and use `common.choose`, you choose

core/ui/public/i18n/en/translation.json
stephdl: sorry it was a duplicate json property, I should have explained it : https://github.com/NethServer/ns8-core/blob/fixSmbValidation/core/ui/public/i18n/en/translation.json#L394

core/ui/public/i18n/en/translation.json
stephdl: https://github.com/search?q=repo%3ANethServer%2Fns8-core%20cluster_placeholder&type=code

we have a function that return all the same, do you want we change things for all ?

```
    selectedProviderHelper() {
      if (this.isBackblazeSelected) {
        return "backblaze_placeholder";
      } else if (this.isAmazonS3Selected) {
        return "aws_placeholder";
      } else if (this.isGenericS3Selected) {
        return "s3_placeholder";
      } else if (this.isAzureSelected) {
        return "azure_placeholder";
      } else if (this.isSambaSelected) {
        return "samba_placeholder";
      } else if (this.isClusterSelected) {
        return "cluster_placeholder";
      } else {
        return "url_placeholder";
      }
    },
```

andre8244: Didn't notice it, it's ok then :v:

andre8244 (approved):

PR #529, opened 2023-11-23
Fix app filter for testing switch

DavidePrincipi:
- Filter the "versions" list, removing the items with a prerelease (testing) tag.
- Skip a package completely, if there is no valid version after the filter is applied.

Fixes https://github.com/NethServer/dev/issues/6777

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
+++ b/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
@@ -47,13 +47,19 @@ def _get_downloaded_logos():
             logos[os.path.basename(app)] = logo
     return logos
 
-def _parse_repository_metadata(repository_name, repository_url, repository_updated, repodata, skip_core_modules = False):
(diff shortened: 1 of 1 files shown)


core/imageroot/usr/local/agent/pypkg/cluster/modules.py
```diff
@@ -47,14 +47,20 @@ def _get_downloaded_logos():
             logos[os.path.basename(app)] = logo
     return logos
 
-def _parse_repository_metadata(repository_name, repository_url, repository_updated, repodata, skip_core_modules = False):
+def _parse_repository_metadata(repository_name, repository_url, repository_updated, repodata, skip_core_modules = False, skip_testing_versions = False):
```
Amygos, line 50: Why default to false? :)

Amygos (approved):

core/imageroot/usr/local/agent/pypkg/cluster/modules.py
Amygos: yes

PR #537, opened 2023-12-14
Fix dirty session module removal

DavidePrincipi:
If there are processes out of the module lingering session the remove-module action fails and leaves the Redis DB inconsistent.

The PR
- fixes the related bug, by sending HUP and TERM signal to all remaining processes of the module.
- implements a `--force` flag for the `remove-module` command: it clears the Redis DB even if remove-module sub-tasks fail

Refs https://github.com/NethServer/dev/issues/6800


gsanchietti (approved):
I'd set the force flag by default when removing the module from the UI

PR #555, opened 2024-01-18
Copy update-core semantics to update-module

DavidePrincipi:
If an update script fails, do not stop the whole module update procedure. Instead continue with the next module update script.

https://github.com/NethServer/ns8-core/blob/43e04becfb40bc973fbc0c873ca0942aeb8c5feb/core/imageroot/var/lib/nethserver/node/actions/update-core/60run_scriptdir#L29

This commit implements for update-module the same semantics of update-core. What do you think?

---
Edit:
Before going in production I want to have the same behavior for both update procedures. As we come from e-smith, I feel more confident with the non-fatal approach.

### DIFF
--- a/core/imageroot/usr/local/agent/actions/update-module/50run_scriptdir
+++ b/core/imageroot/usr/local/agent/actions/update-module/50run_scriptdir
@@ -33,15 +33,22 @@ if [[ ! -d "${AGENT_INSTALL_DIR}"/update-module.d/ ]]; then
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


Amygos:
I don't have any preference, right now, it seems to be the more conservative choice ( if something goes wrong, don't stop all the remaining update processes), but what is the background for the change?

Amygos (approved):

stephdl (approved):

PR #562, opened 2024-01-25
Update core procedure enhancements

DavidePrincipi:
Change the update-core procedure to install the new core image before starting update of core modules. This is important to receive core updates that install fixes to Python libraries or execute Redis updates _before_ any core module is updated.

For this purpose two new directories are defined and can be filled with helper scripts (commands):

- `update-core-pre-modules.d/`
- `update-core-post-modules.d/`

Failed scripts do not abort the update-core action. Such scripts are important because they are executed from the **new core image**, whilst action steps are executed from the **old core image**, at least until step `50update_core`. 

:warning: Bear in mind that the list of action steps to run is generated by the cluster agent once, when the action is invoked: if the step list changes when the new core image is exploded, missing scripts and new scripts are ignored, whilst changed scripts after step `50update_core` are executed from the new image.

The PR also implements a `force` flag in the update-core action, like the similar flag already implemented for update-module. The effect of passing `force:true` is propagated to update-module sub-tasks started to update core modules, too.

See also [Updates](https://github.com/NethServer/ns8-core/blob/04f3e92ca993fb6aaf4cdb5ce7207de7a23ff2cb/docs/core/updates.md) in dev's manual page.

### DIFF
--- a/docs/core/updates.md
+++ b/docs/core/updates.md
@@ -7,40 +7,75 @@ parent: Core
 
 # Core updates
 
-* TOC
-{:toc}
+The core update changes both the core images and the core modules. In the
+first stage cluster nodes update the core components. In the second stage
+the leader node launches the update of individual core modules.
 
-## NS8 updates
+## First stage
 
-Core updates include enhancements and bug fixes for the `core` image. Such image contains: agents, UI and common actions.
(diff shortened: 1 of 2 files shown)


gsanchietti (commented):
:man_shrugging:

PR #565, opened 2024-01-31
Scheduled cluster updates

DavidePrincipi:
- activate `subscription` app repository when a subscription is active, pointing to `subscription.nethserver.com`
- enable automatic scheduled updates when a subscription is active
- update core, core modules, OS (Rocky Linux) and app modules
- implement `apply-updates` cluster script, running on working days
- cluster backup/restore of new Redis key `cluster/apply_updates`
- improve node/update-core, to ensure a non-Semver tag never triggers the image update, unless forced

See NethServer/dev#6835

### DIFF
--- a/core/imageroot/var/lib/nethserver/node/actions/update-core/50update
+++ b/core/imageroot/var/lib/nethserver/node/actions/update-core/50update
@@ -18,14 +18,23 @@ force_update = request.get("force", False)
 
 agent.set_weight(os.path.basename(__file__), 2)
 
+core_env = agent.read_envfile('/etc/nethserver/core.env')
+
+# Fetch the old image information
+with subprocess.Popen(['podman', 'image', 'inspect', core_env['CORE_IMAGE']], stdout=subprocess.PIPE, stderr=sys.stderr) as proc:
+    prev_inspect = json.load(proc.stdout)
+
 try:
-    core_env = agent.read_envfile('/etc/nethserver/core.env')
     aver = semver.parse_version_info(core_env['CORE_IMAGE'].split(':')[-1])
+except ValueError:
+    print(agent.SD_WARNING + f"Current image has no semver tag: {core_env['CORE_IMAGE']}", file=sys.stderr)
+    aver = semver.parse_version_info('9999.9999.9999') # Only a manually forced update can override this big value
+
+try:
     bver = semver.parse_version_info(core_url.split(':')[-1])
 except ValueError:
-    print(agent.SD_WARNING + 'Cannot parse semantic version information from core image tags. Assuming a "minor" update type is occurring', file=sys.stderr)
-    aver = semver.parse_version_info('0.0.0')
-    bver = semver.parse_version_info('0.1.0')
+    print(agent.SD_WARNING + f'New image has no semver tag: {core_url}', file=sys.stderr)
+    bver = semver.parse_version_info('0.0.0') # Only a manually forced update can override this small value
 
 if force_update == False and aver >= bver:
     print('The new core image does not update the current one. Nothing to do.', file=sys.stderr)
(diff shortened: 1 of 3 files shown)


core/imageroot/etc/systemd/system/send-backup.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Store the encrypted cluster backup remotely
 ConditionPathExists=/var/lib/nethserver/cluster/state/backup/passphrase
+After=apply-updates.service
```
gsanchietti, line 4: Why this should be needed? I'd do not bind together the actions, updated info will be send the day after.

core/imageroot/etc/systemd/system/send-inventory.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=Send the system inventory remotely
+After=apply-updates.service
```
gsanchietti, line 3: See below

core/imageroot/usr/local/agent/pypkg/cluster/modules.py
```diff
@@ -98,22 +112,28 @@ def _list_repository_modules(rdb, repository_name, repository_url, skip_core_mod
...
     url = _urljoin(repository_url, "repodata.json")
     try:
-        req = urllib.request.Request(url=url, headers={'User-Agent': 'ns8-downloader'})
-        with urllib.request.urlopen(req) as resp:
-            repodata = resp.read().decode()
+        hsubscription = rdb.hgetall("cluster/subscription")
+        with _get_http_session() as osession:
+            if hsubscription and url.startswith("https://subscription.nethserver.com/"):
+                # Send system_id for HTTP Basic authentication
+                osession.auth = (hsubscription["system_id"], "0000")
```
gsanchietti, line 119: Why not setting also the secret?

core/imageroot/var/lib/nethserver/cluster/bin/apply-updates
```diff
@@ -0,0 +1,51 @@
...
+
+import agent
+import agent.tasks
+import os
+import sys
+
+rdb = agent.redis_connect()
+# Find cluster nodes from the index key
+nodes = list(int(node_id) for node_id in set(rdb.hvals("cluster/module_node")))
+happly_updates = rdb.hgetall("cluster/apply_updates")
```
gsanchietti, line 16: This really looks like a typo ...
```suggestion
apply_updates = rdb.hgetall("cluster/apply_updates")
```

core/imageroot/var/lib/nethserver/cluster/bin/apply-updates
```diff
@@ -0,0 +1,51 @@
...
+        agent_id='cluster',
+        action='update-modules',
+        endpoint="redis://cluster-leader",
+        # disable idle check after core update to avoid a task submission
+        # failure during agent startup:
+        check_idle_time = 0,
+    )
+    if update_modules_response['exit_code'] != 0:
+        print(agent.SD_ERR + "update-modules failed", file=sys.stderr)
+        sys.exit(1)
```
gsanchietti, line 51: Since all updates are bound together, why having 3 different flags?
IMO this is misleading. I'd rather choose 2 different paths:
- if we keep all 3 flags, updates should be independent (eg. if RPM update fails, the app update should still run)
- if we keep one flag, current logic is ok

core/imageroot/var/lib/nethserver/node/actions/update-os/50update_os
```diff
@@ -0,0 +1,35 @@
...
+    "--refresh",
+]
+
+hsubscription = rdb.hgetall("cluster/subscription")
+if hsubscription:
+    # Configure DNF options for HTTP Basic authentication, to retrieve managed
+    # updates
+    dnfargs += [
+        "--setopt=ns-baseos.username=" + hsubscription["system_id"],
+        "--setopt=ns-baseos.password=" + "0000",
```
gsanchietti, line 26: Secret is missing also here?

gsanchietti (changes requested):

core/imageroot/etc/systemd/system/send-backup.service
gsanchietti: This should ensure that the service does not during updates to avoid conflicts.

PR #575, opened 2024-02-13
Fix the restore summary page

DavidePrincipi:
The "Backup date" field is no more available in the cluster backup. It was removed to make the backup archive binary-reproducible.

image

Refs b30633b071a1c2aeb9a6976511a243d3e46ce764


andre8244 (approved):
I'm not sure how Weblate handles the removal of i18n strings from the code for translated languages; I recommend to NOT squash the commit of the PR, even if it's only one.

PR #581, opened 2024-02-21
Configure the agent retry backoff 

DavidePrincipi:
Increase the retry period, to ensure the task outcome is written even if the Redis master is unreachable for a while.

The go-redis library should retry 10 times for a random period of time between 1 and 100 seconds with exponential backoff.

I calculated this range by looking at this code https://github.com/redis/go-redis/blob/cae67723092cac2cb441bc87044ab9edacb2484d/internal/internal.go#L9 and with http://exponentialbackoffcalculator.com/

For a test run: https://go.dev/play/p/xsjwJOZYpk3

Refs NethServer/dev#6854

### DIFF
--- a/core/agent/hevent.go
+++ b/core/agent/hevent.go
@@ -27,6 +27,7 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/NethServer/ns8-core/core/agent/models"
 	"github.com/go-redis/redis/v8"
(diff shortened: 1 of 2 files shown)


Tbaile (commented):
Agree on the retry, will really help in case of network misfires.

However in the worst case scenario we get a retry of almost a minute (if all nine previous attempts fails and on the tenth it succeeds)
This time is enormous, not considering other task steps, to irritate the user and it is not enough to manually solve the issue in case something unplanned happens to the network.

If the usage is merely to wait for redis to restart, this can be a workaround, otherwise I'd keep the time a bit shorter (around 800/1000 ms) and increase the retries to ensure that even long timeouts are still waited.

Tbaile:
And I agree with all the points above, this is why I just gave an opinion :smile: 
If this solves issues with the core reboot and it won't impact the user base, it's just welcome due to the reliability we introduce :rocket:

PR #594, opened 2024-02-29
Display node name or node number inside app selector in "System Logs" view

StefanoScarpetta:
This pull request add the name of the application's node inside the app selector in "System Logs" view, if the node has no name, we will display the node number

image

Ref: NethServer/dev#6870

### DIFF
--- a/core/ui/src/views/SystemLogs.vue
+++ b/core/ui/src/views/SystemLogs.vue
@@ -316,7 +316,7 @@ export default {
               " - " +
               (instance.node_ui_name
                 ? instance.node_ui_name
-                : `Node ${instance.node}`),
+                : this.$t("common.node") + " " + instance.node),
             value: instance.id,
           });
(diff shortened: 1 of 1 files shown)


core/ui/src/views/SystemLogs.vue
```diff
@@ -309,9 +309,14 @@ export default {
...
-              ? instance.ui_name + " (" + instance.id + ")"
-              : instance.id,
+            label:
+              (instance.ui_name
+                ? instance.ui_name + " (" + instance.id + ")"
+                : instance.id) +
+              " - " +
+              (instance.node_ui_name
+                ? instance.node_ui_name
+                : `Node ${instance.node}`),
```
andre8244, line 319: "Node" has to be translated, let's use the corresponding i18n string

andre8244 (approved):

PR #627, opened 2024-05-02
Fix node and cluster state/ dirs permissions

DavidePrincipi:
Ensure state/ dirs are not world-readable as the build process does not correctly set permissions bits.

Refs NethServer/dev#6917


Amygos (approved):
Is there any other non-root user who needs access to read the cluster/node state?

PR #632, opened 2024-05-21
Update SetFqdnModal labels for hostname and domain creation

stephdl:
This pull request updates the labels for the hostname and domain fields in the SetFqdnModal component. The labels were changed to "init.hostname_create" and "init.domain_create" respectively. This change improves the clarity and accuracy of the labels in the user interface.

https://github.com/NethServer/dev/issues/6927

### DIFF
--- a/core/ui/src/components/nodes/SetFqdnModal.vue
+++ b/core/ui/src/components/nodes/SetFqdnModal.vue
@@ -29,7 +29,7 @@
       <cv-form v-else @submit.prevent="setFqdn">
         <template>
           <cv-text-input
-            :label="$t('init.hostname')"
+            :label="$t('init.hostname_create')"
             v-model.trim="hostname"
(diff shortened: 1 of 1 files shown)


stephdl:
either we add a new string in the json file init.domain and init.hostname or we decide to reuse the init.domain_create and init.hostname_create like I did

other way we could decide that this panel cannot use the init.* in the json file and it deserves its own translation

andre8244 (approved):

andre8244:
Found the commit that caused the translation error: https://github.com/NethServer/ns8-core/commit/6cb07ee399b621479bf7e32d42415b6baee96b58

stephdl:
what do you think it could be the better approach @andre8244 ?

andre8244:
> what do you think it could be the better approach @andre8244 ?

This PR is ok, already approved :)

PR #636, opened 2024-05-28
Run the script only at boot

DavidePrincipi:
The unit is pulled in by the Redis unit, and runs before it at boot. However it must run before containers are started, only one time at boot, otherwise the removal of merged/ dirs breaks the overlay filesystem.

Prevent the script to be invoked multiple times, if Redis is restarted.

Refs NethServer/dev#6933


stephdl (approved):

stephdl:
I verified again, debian is still on podman 4.3 :|

PR #646, opened 2024-06-14
App instances count as available applications

DavidePrincipi:
Even if there is no repository enabled I want to see Apps in the Software Center, and Core components too.

This PR changes the list_available() return value, to include those instances that do not belong to any repository (orphans).

Metadata (logo image included) of orphan apps is synthesized from the metadata.json and .png image installed in the cluster leader, if present. Also the image labels are retrieved from the leader image store.

The list_available function is called from many code points. It affects the results of the following actions, without impact on the output format:

- `list-modules`
- `list-installed-modules`
- `list-core-modules`
- `list-updates`
- `update-modules` (invoked also by apply-updates command)

The ordering of available versions and installed versions has been implemented from newest to oldest.

Refs https://github.com/NethServer/dev/issues/6947

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
+++ b/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
@@ -357,47 +357,35 @@ def list_updates(rdb, skip_core_modules = False):
 
     return updates
 
-def list_core_modules():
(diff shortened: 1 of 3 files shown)


stephdl (approved):
LGTM, need to test it in action

PR #651, opened 2024-06-25
Validate leader endpoint URL

DavidePrincipi:
Fail fast if the leader FQDN does not work for any reason.

The most common failure case is a missing DNS record: the URL of the leader is displayed to help to understand the issue.

image

In some secondary cases, the leader FQDN is good and validation succeedes, however the VPN hostname cannot be resolved, or does not work. For these cases:

1. ignore `ns-wireguard` Firewalld service creation error. The service may exist because it was created by a previous join-node attempt
2. check if the VPN hostname is resolved, and abort join-cluster if not. A new join attempt is then possible, once the node has been removed from the cluster.

Refs https://github.com/NethServer/dev/issues/6958

### DIFF
--- a/core/ui/public/i18n/it/translation.json
+++ b/core/ui/public/i18n/it/translation.json
@@ -995,6 +995,7 @@
         "not_a_valid_ipv4_network": "Indirizzo di rete IPv4 non valido",
         "the_join_code_is_not_correctly_encoded": "Il codice di join non ha il formato corretto",
         "cluster_auth_error": "Risposta non valida",
+        "cluster_connection_error": "Connessione al nodo leader all'indirizzo {value} fallita",
         "the_join_code_cannot_be_parsed": "Il codice di join non può essere interpretato",
         "cluster_tls_verify_error": "Il certificato TLS del nodo leader non è valido",
         "the_join_code_can_not_be_decoded": "Il codice di join non può essere decodificato"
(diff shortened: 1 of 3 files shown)


core/ui/public/i18n/it/translation.json
```diff
@@ -995,6 +995,7 @@
         "not_a_valid_ipv4_network": "Indirizzo di rete IPv4 non valido",
         "the_join_code_is_not_correctly_encoded": "Il codice di join non ha il formato corretto",
         "cluster_auth_error": "Risposta non valida",
+        "cluster_connection_error": "Connessione al nodo leader all'indirizzo {value} fallita",
```
andre8244, line 998: Translations should be provided from Weblate webapp; I suspect this change might lock tranlastions.

andre8244 (approved):
UI part LGTM, just a note about Weblate translations

core/imageroot/var/lib/nethserver/cluster/actions/join-cluster/00validate_cluster
```diff
@@ -9,20 +9,21 @@ import json
...
 network = rdb.get('cluster/network')
 
 if network:
     agent.set_status('validation-failed')
-    json.dump([{'field':'url', 'parameter':'url','value': request['url'], 'error':'cluster_network_already_set'}], fp=sys.stdout)
+    json.dump([{'field':'url', 'parameter':'url','value': endpoint_url, 'error':'cluster_network_already_set'}], fp=sys.stdout)
     sys.exit(2)
 
 try:
+    requests.head(endpoint_url, verify=request['tls_verify'], timeout=8.0)
```
gsanchietti, line 26: 8 seconds are a very long time for an HTTP response, I'd lower it to 5 or less

gsanchietti (approved):

PR #657, opened 2024-07-02
Fix redirect to welcome page

andre8244:
This PR solves the following bug:
- https://github.com/NethServer/dev/issues/6967

Other changes:
- Fix a javascript error that appeared on the console while trying to invoke `get-support-session` on the leader node of an uninitialized cluster


andre8244:
> > Click "Create cluster"
> 
> I got the same bug if I click the Join node button. Does the fix cover this usage case too?

Yes

edospadoni (approved):

PR #667, opened 2024-07-12
Add image core and app version constraint checks

DavidePrincipi:
Implement module image labels

- `org.nethserver.min-from`
- `org.nethserver.min-core`

See changes to `modules.md`.

Implementation summary:
- core comparison for `min-core` check is based on the node version where the app is installed
- update for a given app instance is compared with `min-from` label [^1]
- Redis HASH `cluster/module_node` is authoritative to get the list of installed modules and cluster node IDs
- `list_installed_core()` has been merged within list_core_modules()

Affected actions:

- list-modules
- list-installed-modules
- list-core-modules
- list-updates
- update-modules
- update-module
- update-core

Fixed install-core.sh

Refs https://github.com/NethServer/dev/issues/6956

[^1]: list_available() already returns in versions[0] the higher repository semver tag of the app


Tbaile (approved):
Looks much like what discussed today

compgeniuses:
Wow, this will solve so many problem. looking forward to this in a future release

PR #670, opened 2024-07-12
UI - Cloud Log Manager modals shows cluster id

tommaso-ascani:
Implemented inline notification to show cloud log manager cluster id when it's present

[Nethserver/dev#6932](https://github.com/NethServer/dev/issues/6932)


core/ui/src/components/settings/CloudLogManagerConfigureModal.vue
```diff
@@ -17,6 +17,12 @@
     }}</template>
     <template slot="content">
       <cv-form v-if="subscription">
+        <NsInlineNotification
+          v-if="this.cluster_id != ''"
```
andre8244, line 21: - Let's use `v-if="clusterId"`. The condition `this.cluster_id != ''` should be fine but might be too specific: what if `cluster_id == undefined` instead of empty string?
- `this` is not needed inside templates

core/ui/src/components/settings/CloudLogManagerConfigureModal.vue
```diff
@@ -148,6 +154,7 @@ export default {
       toggleEnabled: false,
       address: "",
       tenant: "",
+      cluster_id: "",
```
andre8244, line 157:
```suggestion
      clusterId: "",
```
Let's use camel case for variables

core/ui/src/components/settings/CloudLogManagerConfigureModal.vue
```diff
@@ -202,6 +209,13 @@ export default {
         this.tenant = this.configuration.tenant;
       }
 
+      if (!this.configuration.cluster_id == "") {
+        this.cluster_id =
```
andre8244, line 213: This string is not internationalized. Let's use a i18 string with a parameter, i.e. `This cluster is identified as {clusterId} in the Cloud Log Manager UI`

andre8244 (changes requested):

core/ui/src/components/settings/CloudLogManagerConfigureModal.vue
```diff
@@ -202,6 +213,10 @@ export default {
         this.tenant = this.configuration.tenant;
       }
 
+      if (!this.configuration.cluster_id == "") {
```
andre8244, line 216:
```suggestion
      if (this.configuration.cluster_id) {
```

tommaso-ascani:
image
image
image

andre8244 (approved):

PR #672, opened 2024-07-15
chore: update header comments

andre8244:
Update year in header comments


andre8244:
> 😲 Wait, we never updated Copyright year massively!

We did it last year :thinking: No problem for me anyway, one less thing to do :man_shrugging:

gsanchietti:
I agree, this is not really needed!

andre8244:
No need to update header comments :relieved:

PR #673, opened 2024-07-15
Automatic updates see different metadata

DavidePrincipi:
This is the client-side implementation of different NS8 metadata repository views.

The user can always see the latest available metadata from the Software Center, while the nightly update cronjob, which runs if Subscription is active, sees the managed metadata contents.

Refs NethServer/dev#6970


Tbaile (approved):
While I'd suggest using a parameter instead of a header for the subscription repositories, everything else checks out

PR #674, opened 2024-07-16
fix: cards in grids now have the same height

andre8244:
Cards in the same row of a card grid should have the same height. This PR provides the required CSS rules and applies the fix to some card grids.


Tbaile (approved):
Pure style review, haven't noticed anything that causes issues

PR #691, opened 2024-08-05
Add 'node_id' attribute to core modules

stephdl:
This pull request adds the 'node_id' attribute to the core modules in the codebase. The 'node_id' attribute is used to identify the node where the module is installed. This change ensures that the 'node_id' is included in the core module list and is displayed in the software center UI.

https://github.com/NethServer/dev/issues/6989

image
image
image
image


andre8244 (commented):
Is it possible to add node label, if present? In NS8 UI, a node is usually named:
- "My first node (Node 1)" if node label is present
- "Node 1" if node label is not present

If this change makes sense, I also suggest to rename "Node ID" column to to "Node"

stephdl:
do you mean to display in the column instead of `1` rename to `Node 1`

andre8244:
> do you mean to display in the column instead of `1` rename to `Node 1` @andre8244 ?

Yes, or display "My node (Node 1)" if the node has a label "My node" set

andre8244:
> > do you mean to display in the column instead of `1` rename to `Node 1` @andre8244 ?
> 
> Yes, or display "My node (Node 1)" if the node has a label "My node" set

You can use [getNodeLabel()](https://github.com/NethServer/ns8-ui-lib/blob/main/src/lib-mixins/util.js#L287) function from ns8-ui-lib

stephdl:
I need to share with @DavidePrincipi, I am not sure it is in the scope, the goal is to sort by node ID, and any verbose label could break it
What do you think davidep

andre8244:
> I need to share with @DavidePrincipi, I am not sure it is in the scope, the goal is to sort by node ID, and any verbose label could break it What do you think davidep

NsDataTable component supports custom sorting with `customSortTable` prop, you can find an example here: https://github.com/NethServer/ns8-mail/blob/main/ui/src/components/mailboxes/UserMailboxes.vue#L15

andre8244 (approved):

PR #692, opened 2024-08-05
Ensure environemnt var names starts with non-digit

DavidePrincipi:
Rules for environment var names are stricter than container image names: they cannot start with a digit. If an image name starts with a digit prepend "I" to avoid Systemd and shell errors.

Refs https://github.com/NethServer/dev/issues/6990


Tbaile (approved):
To whoever starts their own application name with a digit, I will find you.

PR #694, opened 2024-08-06
Fix VPN routing error

DavidePrincipi:
If two nodes are in distinct private LAN networks, avoid bad VPN routing configuration: never push an endpoint IP inside the VPN itself. 

The PR also clean up the `destmap` hash from the unused key corresponding to the local node_id (self_id). Neither Wireguard nor the IP routing table need an entry for the local node.

Refs NethServer/dev#6976

### DIFF
--- a/core/imageroot/usr/local/sbin/apply-vpn-routes
+++ b/core/imageroot/usr/local/sbin/apply-vpn-routes
@@ -52,21 +52,20 @@ for node_id in peers:
         destmap[leader_id].update(peers[node_id]['destinations'].split())
         destmap[leader_id].add(peers[node_id]['ip_address'])
 
-def get_wgset_endpoint_clause(endpoint):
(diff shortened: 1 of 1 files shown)


core/imageroot/usr/local/sbin/apply-vpn-routes
```diff
@@ -85,19 +82,21 @@ for iface in ipaddr_reply:
 errors = 0
 valid_destinations = []
 for node_id in destmap:
+    peer_host, peer_port = resolve_endpoint(peers[node_id].get('endpoint', ''))
+    allowed_ips = destmap[node_id] - {peer_host}
```
Tbaile, line 86: Bit unsure about this, do you wrap the `peer_host` in a `dict` to run a difference between `dict`s?

Tbaile (approved):

PR #702, opened 2024-09-06
fix: infinite scrolling in sw center and app drawer

andre8244:
Fix page size of infinite scrolling in software center and app drawer: on larger screens infinite scrolling didn't work.


Tbaile (approved):
:tractor:

PR #782, opened 2025-01-14
Fix wrong permissions after core update

DavidePrincipi:
- Avoid concurrent `tar` runs, which leave wrong dir permissions and generate various system services failures (e.g. api-server).
- Ensure node IDs are not duplicated.

Refs NethServer/dev#7250


Amygos:
> * Ensure node IDs are not duplicated.

Is it related to the main issue?

Amygos (approved):

PR #783, opened 2025-01-15
Extended list-user-domains response

DavidePrincipi:
Optionally retrieve an universal unique identifier of the LDAP DB, by looking at the base DN entryUUID or GUID attribute. The implementation should work with many LDAP implementations.

A global identifier of the LDAP DB is necessary to validate the NS8 external LDAP user domain during NS7 migration.

Refs NethServer/dev#7222


gsanchietti (approved):
I'm not really sure that such entries are always present.
While it's ok to retrieve them, I'd not use them for strict validation.

PR #789, opened 2025-01-23
feat(runagent): list running agent IDs

DavidePrincipi:
Implement a runagent option `--list-modules` (`-l`) that prints the names of agents running on the local node.

This option is used by Bash completion to show the agent names by pressing TAB after `-m`. E.g.

    runagent -m t<TAB> => runagent -m traefik1
    
Discussion https://mattermost.nethesis.it/nethesis/pl/dofpmuyytpbcjyhbd3zzjb5uto


gsanchietti (commented):
:shrug: 
IMO it does not add so much value (just new code to maintain), but fell free to merge if it works.

PR #807, opened 2025-02-13
fix(test): new get-facts format

DavidePrincipi:
Run the test on the print-phonehome output instead of get-facts.

Refs https://github.com/NethServer/ns8-core/pull/784


gsanchietti (commented):
:shrug:

PR #809, opened 2025-02-14
fix(ldap): Update domain configuration and reload event

stephdl:
Refine the domain configuration logic for internal Active Directory schemas and emit a reload event after adding an external provider.

https://github.com/NethServer/dev/issues/7315

### DIFF
--- a/core/ui/src/views/DomainConfiguration.vue
+++ b/core/ui/src/views/DomainConfiguration.vue
@@ -1049,7 +1049,7 @@ export default {
       }
       this.currentProvider = provider;
 
-      if (this.domain.schema === "ad") {
+      if (this.domain.schema === "ad" && this.domain.location == "internal") {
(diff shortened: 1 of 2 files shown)


core/ui/src/views/DomainConfiguration.vue
```diff
@@ -1049,7 +1050,7 @@ export default {
       }
       this.currentProvider = provider;
 
-      if (this.domain.schema === "ad") {
+      if (this.domain.schema === "ad" && this.domain.location == "internal") {
```
andre8244, line 1053: I don't understand this change, external providers shouldn't have delete button and shouldn't get to this point of the code :thinking:

stephdl: simple, we use the internal removal action for an external ldap provider

andre8244 (approved):

PR #817, opened 2025-02-20
fix(create-cluster): use Traefik config of workers

DavidePrincipi:
Traefik API connections originate only from the localhost, access from cluster VPN is not needed.

The cluster VPN is not automatically inserted in Traefik configuration of worker nodes. As result there is a configuration difference between the original cluster leader and other nodes. This commit wants to remove this conflicting configuration.

Refs NethServer/dev#7312
Reverts 7c90ee261983df108715deef714b765360f45012


gsanchietti (approved):
No problem to remove, but since it does not hurt, why not keeping it to avoid regressions?

PR #824, opened 2025-03-07
Implement get-certificate helper for apps

DavidePrincipi:
This PR expands the agent Python library and adds a new command to improve the management of TLS certificates in NS8 applications.

- With Traefik v3 a new event is notified when a certificate changes, `certificate-changed`. 
- The application event handler can check if the event is relevant with function `agent.certificate_event_matches()`
- The application can obtain the certificate and key with `get-certificate` command.
- Wildcard match is supported.
- The `get-certificate` command checks if the obtained certificate and key differ from the current certificate.
- If the remote task fails, `get-certificate` uses the existing files. If they are not present a self-signed certificate is generated, to prevent service startup failures.

Refs  NethServer/dev#7004

See also: https://github.com/NethServer/ns8-traefik/pull/89

### DIFF
--- a/core/imageroot/usr/local/agent/bin/get-certificate
+++ b/core/imageroot/usr/local/agent/bin/get-certificate
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+#
+# Copyright (C) 2025 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import argparse
+import sys
+import os
+import agent
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.asymmetric import rsa
+from datetime import datetime, timedelta
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="""
+        Retrieves and saves a TLS certificate and private key. If the certificate retrieval fails
+        and no existing certificate is found, a self-signed certificate is generated.
+
+        Exit codes:
+            0 - Certificate successfully obtained or updated.
+            1 - Generic error (e.g., file I/O failure, retrieval failure with existing files).
+            3 - Certificate remains unchanged.
+        """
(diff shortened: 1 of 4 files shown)


gsanchietti (approved):

core/imageroot/usr/local/agent/bin/get-certificate
```diff
@@ -0,0 +1,125 @@
...
+        public_exponent=65537,
+        key_size=2048,
+    )
+    subject = x509.Name([
+        x509.NameAttribute(NameOID.COMMON_NAME, fqdn)
+    ])
+    cert = (
+        x509.CertificateBuilder()
+        .subject_name(subject)
+        .issuer_name(subject)
```
gsanchietti, line 109:
```suggestion
        .issuer_name("get-certificate")
```

gsanchietti (approved):

core/imageroot/usr/local/agent/bin/get-certificate
```diff
@@ -0,0 +1,125 @@
...
+            sys.exit(1)
+    return ""
+
+def atomic_write(file_path: str, content: str) -> None:
+    """Writes content to a temporary file and renames it to avoid corruption."""
+    temp_path = f"{file_path}.tmp"
+    try:
+        with open(temp_path, "w") as temp_file:
+            temp_file.write(content)
+        os.replace(temp_path, file_path)
```
Amygos, line 92: Only a note, the atomic operation is valid only if the temp cert file and the old cert file are in the same filesystem, otherwise the certificate will be copied.

Amygos (approved):

core/imageroot/usr/local/agent/bin/get-certificate
Amygos: It can happen if the module is installed on a directory mounted from an external volume.

Amygos: Yep, I missed that.

PR #843, opened 2025-03-26
fix(ui): HTTP route creation

DavidePrincipi:
The creation form fields are hidden except for "ip allow list". Invert the default user_created default value to show them.

Refs NethServer/dev#7362


gsanchietti (approved):
Please test the edit also of automatic routes.

PR #850, opened 2025-04-09
feat(add-public-service): implement ports addition

DavidePrincipi:
Allow multiple invocations of add-public-service with or without ports replacement. This new behavior simplifies the addition of new services to existing applications, like Samba WSDD.

Implementation note: `firewall-cmd` does not allow to use both `--add-port` and `--remove-port` in the same invocation, so I split them in two command calls.

Refs NethServer/dev#7372


gsanchietti (commented):
This is way long too complicated.
We can achieve the same using existing functions:
```
agent.remove_public_service(os.environ['MODULE_ID'])
agent.add_public_service(os.environ['MODULE_ID'], ["9010/tcp", "9011/tcp"])
```

PR #852, opened 2025-04-09
fix: add more specific no_version_reason

DavidePrincipi:
If no stable app version is available from a repository return a specific error type to avoid confusion with the org.nethserver.min-core compatibility check.

Refs NethServer/dev#7228

### DIFF
--- a/core/ui/public/i18n/en/translation.json
+++ b/core/ui/public/i18n/en/translation.json
@@ -968,7 +968,7 @@
         "no_node_eligible_for_instance_cloning": "No node is eligible for instance cloning",
         "no_node_eligible_for_instance_migration": "No node is eligible for instance migration",
         "no_version_reason": {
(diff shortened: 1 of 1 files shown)


core/ui/public/i18n/en/translation.json
```diff
@@ -968,6 +968,7 @@
         "no_node_eligible_for_instance_cloning": "No node is eligible for instance cloning",
         "no_node_eligible_for_instance_migration": "No node is eligible for instance migration",
         "no_version_reason": {
+            "no_app_version": "No stable release of {app_name} is available in the \"{app_repository}\" repository",
```
andre8244, line 971:
```suggestion
            "no_app_version": "No stable release of {app_name} is available in the '{app_repository}' repository",
```
We usually use single quotes inside labels

andre8244 (approved):
Just a minor label improvement

PR #857, opened 2025-04-14
feat: set Redis TTL for task /context

DavidePrincipi:
The context key is created in Redis with a short TTL and refreshed until the task runs together with each event heartbeat. On task completion the TTL is set to the previous, 8-hours long, value.

The task is displayed as running as long as the context key exists and the progress events are sent through web socket.

If the task heartbeat is lost the context key expires and the task is no longer displayed in the UI.

Refs NethServer/dev#7319


gsanchietti (approved):
:shrug:

PR #858, opened 2025-04-14
feat(add-node): worker version sanity check

DavidePrincipi:
Check if worker's major.minor version numbers are the same of the leader's Core version. Special exception for NS7 migration tool that is always allowed to join the cluster.

The leader version is returned also in the action's output as additional information.

Refs NethServer/dev#7376

### DIFF
--- a/core/ui/public/i18n/en/translation.json
+++ b/core/ui/public/i18n/en/translation.json
@@ -207,6 +207,8 @@
         "the_join_code_can_not_be_decoded": "Join code cannot be decoded",
         "the_join_code_is_not_correctly_encoded": "Join code is not correctly encoded",
         "the_join_code_cannot_be_parsed": "Join code cannot be parsed",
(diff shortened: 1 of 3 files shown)


core/ui/public/i18n/en/translation.json
```diff
@@ -207,6 +207,8 @@
         "the_join_code_can_not_be_decoded": "Join code cannot be decoded",
         "the_join_code_is_not_correctly_encoded": "Join code is not correctly encoded",
         "the_join_code_cannot_be_parsed": "Join code cannot be parsed",
+        "core_version_mismatch": "To join this cluster core version {value} is required",
```
gsanchietti, line 210: I'd suggest a solution here because is a blocking condition.
Maybe something like:
```
To join this cluster core version {value} is required. Make sure to update the cluster leader or reinstall the worker with the latest available image.
```

gsanchietti (approved):

PR #869, opened 2025-04-24
Add script to clean up all restic images in NethServer modules

stephdl:
Introduce a script that removes all restic images in NethServer modules after a core update

https://github.com/NethServer/dev/issues/7391


stephdl:
it should be good @DavidePrincipi

PR #885, opened 2025-05-26
feat(ui): enable set-certificate conflict check

DavidePrincipi:
Prevent from adding a FQDN that is already configured by an HTTP route with Let's Encrypt flag.

Refs NethServer/dev#7383

### DIFF
--- a/core/ui/public/i18n/en/translation.json
+++ b/core/ui/public/i18n/en/translation.json
@@ -799,7 +799,7 @@
         "node": "Node",
         "request_certificate_for_fqdn": "Request certificate for {fqdn}",
         "fqdn_already_exists": "FQDN already exists",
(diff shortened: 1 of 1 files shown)


core/ui/public/i18n/en/translation.json
```diff
@@ -799,6 +799,7 @@
         "node": "Node",
         "request_certificate_for_fqdn": "Request certificate for {fqdn}",
         "fqdn_already_exists": "FQDN already exists",
+        "fqdn_conflicts_with_route": "A certificate for this FQDN is already requested by an HTTP route",
```
andre8244, line 802:
```suggestion
        "fqdn_conflicts_with_route": "A certificate for this FQDN is already requested by an HTTP route. No need to request again the certificate",
```

PR #932, opened 2025-09-10
Enforce password expiration policy in user management

stephdl:
Implement password expiration policy enforcement in the user modal and update the display logic in the DomainUsers component to reflect this policy.

https://github.com/NethServer/dev/issues/7503


stephdl:
@DavidePrincipi  what do you think, can you approve the translation string ?

PR #947, opened 2025-10-03
feat(alloy): security label for auth messages

DavidePrincipi:
Add the "security" category label to auth messages from api-server.

Refs  NethServer/dev#7667


gsanchietti (approved):
🤷

PR #949, opened 2025-10-07
fix(redis): security fixes

DavidePrincipi:
Upstream fixes, mainly for CVEs:
- 2025-49844
- 2025-21605


stephdl (commented):
do we update the core or we fix only new installation ?

gsanchietti (approved):

PR #1167, opened 2026-04-29
Update visibility condition for password expiration policy toggle

stephdl:
Refine the condition for displaying the password expiration policy toggle to ensure it reflects the correct state based on the policy enforcement.

https://github.com/NethServer/dev/issues/7981

### DIFF
--- a/core/ui/src/components/domains/DomainUsers.vue
+++ b/core/ui/src/components/domains/DomainUsers.vue
@@ -299,6 +299,12 @@ export default {
       this.listDomainUsers();
     }
   },
+  mounted() {
+    this.$root.$on("password-policy-updated", this.listDomainUsers);
+  },
+  beforeDestroy() {
+    this.$root.$off("password-policy-updated", this.listDomainUsers);
+  },
   methods: {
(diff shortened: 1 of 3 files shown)


core/ui/src/components/domains/DomainUsers.vue
```diff
@@ -299,6 +299,12 @@ export default {
       this.listDomainUsers();
     }
   },
+  mounted() {
+    this.$root.$on("password-policy-updated", this.listDomainUsers);
```
andre8244, line 303: Where does this event come from?

stephdl: in domain configuration when we change the policy and we go back to the tab of the users, the table is not updated in the background and we are not able to see for example that the password is time limited or unlimited if the policy has changed, this fix does it

see the line in DomainConfiguration.vue in this PR

andre8244: Sorry, I missed it 😅

andre8244 (approved):

PR #1175, opened 2026-05-13
feat(backup): route all backup traffic through rclone-gateway

### LINKED ISSUE #7814: Centralized Backup Orchestration and Rclone-based Gateway
This proposal revises the current NS8 backup architecture by moving scheduling and orchestration from individual applications to the node level, with the goal of reducing resource usage, simplifying destination support, and improving consistency.

Current architecture
Each NS8 application runs Restic independently and pushes backups directly to a configured Destination (e.g. S3 bucket), according to one or more Schedules implemented as app-specific systemd timer units.

Current limitations

* All application timers typically trigger at the same time on a node, causing multiple backups to run in parallel and leading to high memory and CPU consumption.
* Adding support for additional cloud backup providers requires code changes. At present, Restic is used directly for AWS, Azure, and B2, while SMB and generic S3 destinations rely on Restic behind rclone serve, resulting in duplicated logic and inconsistent handling.

Proposed architecture

* Backup Schedules are implemented as systemd timers at the node-agent level. When a schedule triggers, the node agent runs application backups sequentially on that node, impersonating each app as needed (e.g. via `runagent -m appX`) to keep resource usage under control.
* All backup destinations are accessed through Rclone, with support for importing a raw `rclone.conf` file to allow flexible and extensible backend configuration without code changes.
* Applications send backup data through a node-level gateway implemented with `rclone serve rest`. Applications authenticate to the gateway and are granted append-only Restic access to their backup destinations.
* After backup execution, the node agent is responsible for running Restic prune operations, enforcing the configured retention policy and cleaning up old snapshots centrally.

This change centralizes backup scheduling and retention logic, reduces peak resource usage on nodes, and standardizes destination handling across all applications.

### LINKED ISSUE #7854: Validate write permissions on backup destination
When a new backup destination is added and its settings are validated, try to create a remote file to ensure write permissions are granted.

Refs https://community.nethserver.org/t/ns8-backup-failing-on-idrive-e2-s3/24936/3

Thanks to **Dan**

### LINKED ISSUE #7883: Restore backup schedule within disaster recovery
After disaster recovery, restored applications are already scheduled for backup as they were in the original cluster.

**Proposed solution**

- Add relevant backup schedule information to cluster backup, e.g. module UUIDs

**Alternative solutions**

- n/a

DavidePrincipi:
Modules no longer access backup destinations directly. A per-node `rclone-gateway` service (HAProxy + Rclone) acts as an authenticated HTTP proxy: nodes get full WebDAV and Restic REST access, while applications are restricted to their own repositories.

This removes secret credentials from module environments, centralizes destination configuration in private Redis keys, and enables node-level backup orchestration with `run-backup`.

Main changes:

- **rclone-gateway service** replaces `rclone-webdav`, proxying Restic REST and WebDAV backends with Redis-based auth
- **Secrets isolation** — rclone configs and Restic passwords stored in `private/nodes/*` and `private/agents/*`, inaccessible to modules
- **Node-level orchestration** — `run-backup` iterates modules, retries locked backups, writes Prometheus metrics, uploads metadata and cluster backup via WebDAV
- **Systemd scheduling** — `schedule-backup` manages transient timers from `cluster/backup/*` definitions
- **API hardening** — obfuscate `*key` and `*pass` params in audit log
- **Migration** — node Redis ACLs and backup keys updated for existing installations

Refs NethServer/dev#7814


core/imageroot/etc/systemd/system/rclone-gateway.service
```diff
@@ -0,0 +1,50 @@
...
+    --cgroups=no-conmon \
+    --detach \
+    --init \
+    --log-opt=tag=%N \
+    --replace --name=%N \
+    --network=host \
+    --volume=./rclone:/etc/rclone:ro,Z \
+    --volume=./haproxy:/etc/haproxy:ro,Z \
+    --volume=${BACKUP_VOLUME}:/srv/repo:z \
+    --mount=type=tmpfs,tmpfs-size=10M,destination=/var/lib/rclone,chown=true \
```
gsanchietti: Is this special option really needed?

core/imageroot/usr/local/agent/bin/module-backup
```diff
@@ -21,49 +21,34 @@
...
-import time
-import atexit
 
-backup_status = {'total_size': 0, 'total_file_count': 0, 'snapshots_count': 0, 'errors': 0, 'start': int(time.time())}
+EXIT_ALREADY_RUNNING = 3
 
 backup_id = int(sys.argv[1])
-rdb = agent.redis_connect(host='127.0.0.1') # Connect to local replica
+
+lock_file = open(os.environ['AGENT_STATE_DIR'] + f"/.module-backup.lock", "w")
```
gsanchietti, line 34: Not a fun of locks.
Can be avoid?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
```diff
@@ -40,3 +54,371 @@ def get_default_backup_repository_name(provider, url, rid=""):
...
+
+# Same key hardcoded in rclone source
+_CRYPT_KEY = bytes([
+    0x9c, 0x93, 0x5b, 0x48, 0x73, 0x0a, 0x55, 0x4d,
+    0x6b, 0xfd, 0x7c, 0x63, 0xc8, 0x86, 0xa9, 0x2b,
+    0xd3, 0x90, 0x19, 0x8e, 0xb8, 0x12, 0x8a, 0xfb,
+    0xf4, 0xde, 0x16, 0x2b, 0x8b, 0x95, 0xf6, 0x38,
+])
+
+_AES_BLOCK_SIZE = 16  # aes.BlockSize in Go
```
gsanchietti, line 337: This is scary because relies on rclone internals.
Can we avoid it?

core/imageroot/var/lib/nethserver/cluster/actions/remove-backup-repository/50remove_backup_repository
```diff
@@ -42,32 +42,26 @@ trx = rdb.pipeline()
 #
 # Drop related backup objects
 #
-ptasks = []
+related_backups = []
+# XXX use an index instead of scan
```
gsanchietti: Should we remove this comment?

core/imageroot/var/lib/nethserver/node/actions/run-backup/50run_backup
```diff
@@ -0,0 +1,18 @@
...
+import sys
+import json
+import agent
+
+request = json.load(sys.stdin)
+bid = request['id']
+
+proc = agent.run_helper("run-backup", f"--backup={bid}")
+if proc.returncode != 0:
+    sys.exit(1)
```
gsanchietti, line 18: Why not returning the real exit code?
```suggestion
return proc.returncode
```

core/imageroot/var/lib/nethserver/node/install-finalize.sh
```diff
@@ -97,7 +97,11 @@ EOF
 ) | redis-cli
 
 echo "Start API server and core agents:"
-systemctl enable --now api-server.service agent@cluster.service agent@node.service rclone-webdav.service
+systemctl enable --now \
+    api-server.service \
+    agent@cluster.service \
+    agent@node.service \
+    # end of service list
```
gsanchietti, line 104: This is a but ugly, especially the last line with a comment.

```suggestion
systemctl enable --now 
    api-server.service \
    agent@cluster.service \
    agent@node.service
```

core/imageroot/var/lib/nethserver/node/bin/run-backup
```diff
@@ -0,0 +1,264 @@
...
+    content = (
+        '# HELP node_backup_status Status of the backup (0 = failure, 1 = success, 2 = conflict, -1 = unknown)\n'
+        '# TYPE node_backup_status gauge\n'
+        f'node_backup_status{{id="{backup_id}",name="{prometheus_escape_label_value(backup_name)}"}} {status}\n'
+    )
+    os.makedirs(PROM_DIR, exist_ok=True)
+    tmp_file = f"{output_file}-{os.getpid()}.tmp"
+    with open(tmp_file, "w") as f:
+        f.write(content)
+    os.rename(tmp_file, output_file)
```
gsanchietti, line 49: Is this backward compatibile?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
gsanchietti: Suggestion use: https://rclone.org/commands/rclone_obscure/

core/imageroot/var/lib/nethserver/node/actions/run-backup/50run_backup
gsanchietti: Not needed

core/imageroot/var/lib/nethserver/node/bin/run-backup
gsanchietti: Yes it is

core/imageroot/var/lib/nethserver/node/events/backup-schedule-changed/20purge_backup_prom
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+#
+# Copyright (C) 2024 Nethesis S.r.l.
```
gsanchietti:
```suggestion
# Copyright (C) 2026 Nethesis S.r.l.
```

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
DavidePrincipi: I obtained this patch, but I think the relying on Podman at runtime - especially during the core update - can be fragile. I'm confident Rclone will not change its hardcoded to avoid breaking existing config files.
https://gist.github.com/DavidePrincipi/7406ded6919a476fd17b7683d54f4678

gsanchietti: :crossed_fingers:

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
```diff
@@ -40,3 +54,371 @@ def get_default_backup_repository_name(provider, url, rid=""):
 def validate_schedule(schedule):
     """Check the given schedule string is good for Systemd timers"""
     return subprocess.run(['/usr/bin/systemd-analyze', 'calendar', schedule], stdout=subprocess.DEVNULL).returncode != 0
+
+def run_rclone(rclone_args, temp_config=None, podman_args=[], **runargs):
```
stephdl, line 58: `podman_args=[]` is a mutable default that we then mutate in place (`podman_args += [...]` and `podman_args.append("-i")`). Wouldn't successive calls that don't pass `podman_args` accumulate flags from previous invocations — e.g. a stale `RCLONE_CONFIG` or a leftover `-i` leaking across calls? Should we use `podman_args=None` and build a fresh list inside the function instead?

core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -212,68 +212,70 @@ def run_helper(*args, log_command=True, **kwargs):
...
+        restic_password = rdb.hget('private/agents/backup_destination/restic_password', repository)
+    except redis.exceptions.NoPermissionError:
+        # Warn app devs to fix the connection privileges of rdb argument:
+        warnings.warn("Fix your agent.prepare_restic_command() or agent.run_restic() invocation by passing a privileged Redis connection (e.g.: \"redis_connect(privileged=True)\").", stacklevel=3)
+        rdbrw = redis_connect(privileged=True)
+        restic_password = rdbrw.hget('private/agents/backup_destination/restic_password', repository)
+
+    # Build the environment to run Restic against the given repository+repo_path
+    restic_env = {
+        "RESTIC_PASSWORD": restic_password,
```
stephdl, line 250: `rdb.hget(...)` (lines 242 / 247) returns `None` when the field is missing, so `RESTIC_PASSWORD` can end up `None` here and only blow up later inside `subprocess` with an opaque `TypeError` far from the cause. Should we `assert_exp(restic_password)` after both the normal and the privileged-retry reads, the same way we already `assert_exp(len(orepo) > 0)` for the repository?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
```diff
@@ -40,3 +54,371 @@ def get_default_backup_repository_name(provider, url, rid=""):
...
+    url = f"{base_url.rstrip('/')}/{path.lstrip('/')}"
+    r = session.get(url)
+    r.raise_for_status()
+    return r.json()
+
+def webdav_write_file(session, base_url, path, data: bytes, content_type="application/octet-stream"):
+    url = f"{base_url.rstrip('/')}/{path.lstrip('/')}"
+    # Make sure parent exists:
+    parent = '/'.join(url.split('/')[:-1])
+    session.request("MKCOL", parent)
```
stephdl, line 424: `MKCOL` is non-recursive and here we only create a single parent level, and the response isn't checked. If any intermediate collection is missing, won't this fail (409) and then surface as a confusing `PUT` error instead? Should we create each path segment in turn and check the MKCOL status so a real failure is reported clearly?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
```diff
@@ -40,3 +54,371 @@ def get_default_backup_repository_name(provider, url, rid=""):
...
+    elif uschema == 's3':
+        if params['aws_secret_access_key'] == "":
+            secret = fbcfg.get('secret_access_key')
+        else:
+            secret = params['aws_secret_access_key']
+        rclone_conf = (
+            f"type = s3\n"
+            f"env_auth = true\n"
+            f"access_key_id = {params['aws_access_key_id']}\n"
+            f"secret_access_key = {secret}\n"
```
stephdl, line 181: Secrets are interpolated raw into the INI text here (this applies to `secret_access_key`, the b2 `key`, the azure `key`, hosts, regions — SMB `pass` is fine since it's obscured). Since this string gets parsed back by `configparser` later, couldn't a value containing a newline or a `[` inject an extra rclone key or even a new section? Would it be safer to build the config with `configparser` / write it out, instead of f-string concatenation, so values are escaped consistently with `_normalize_rclone_conf`?

core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -212,68 +212,70 @@ def run_helper(*args, log_command=True, **kwargs):
...
-        restic_args.insert(0, "--option=rclone.program=/usr/local/bin/rclone-wrapper")
-    elif uschema == 'webdav' and orepo['provider'] == 'cluster':
-        ourl = urlparse(upath)
-        restic_env["RESTIC_REPOSITORY"] = 'rclone::webdav:' + ourl.path.rstrip("/") + "/" + repo_path
-        restic_env["RCLONE_WEBDAV_URL"] = ourl.scheme + '://' + ourl.netloc
+    if os.environ['REDIS_USER'] == 'cluster':
+        # cluster credentials are not known to rclone-gateway, use node
+        # creds instead.
+        node_env = read_envfile('/var/lib/nethserver/node/state/agent.env')
+        http_auth = (node_env['REDIS_USER'], node_env['REDIS_PASSWORD'])
```
stephdl: `http_auth` is assigned here and again on line 237, but I can't find it being used anywhere afterwards in `prepare_restic_command` — only `restic_rest_username` / `restic_rest_password` are. Is this leftover from an earlier design that we can drop, or was the gateway auth meant to actually use `http_auth` somewhere (i.e. a wiring bug)?

core/imageroot/usr/local/agent/pypkg/agent/__init__.py
```diff
@@ -212,68 +212,70 @@ def run_helper(*args, log_command=True, **kwargs):
...
 def run_restic(rdb, repository, repo_path, podman_args, restic_args, progress_callback=None, **kwargs):
     podman_cmd, restic_env = prepare_restic_command(rdb, repository, repo_path, podman_args, restic_args)
 
     penv = os.environ.copy()
     penv.update(restic_env)
-    if os.getenv('DEBUG', False):
-        print(*([f"{k}={v}" for k,v in restic_env.items()] + podman_cmd), file=sys.stderr)
-    else:
+    if os.getenv('DEBUG') == "1":
         print("restic", *restic_args, file=sys.stderr)
```
stephdl, line 278: Dropping the old DEBUG branch that dumped the full env (with secrets) is a good fix. But the old `else` always logged `restic <args>` unconditionally — now with `DEBUG` unset nothing is logged at all. Was removing that default breadcrumb intentional, or should we keep an unconditional `print("restic", *restic_args)` and only gate the verbose output behind `DEBUG == "1"`?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
```diff
@@ -40,3 +54,371 @@ def get_default_backup_repository_name(provider, url, rid=""):
...
+            result['azure_account_key'] = ""
+        else:
+            result['azure_account_key'] = dsection.get('key', '')
+    elif rtype == 'smb':
+        result['smb_host'] = dsection.get('host', '')
+        result['smb_user'] = dsection.get('user', '')
+        if hide_secrets:
+            result['smb_pass'] = ""
+        else:
+            result['smb_pass'] = rclone_reveal(dsection.get('pass'))
```
stephdl: `dsection.get('pass')` returns `None` when the field is absent, and `rclone_reveal(None)` would do `None + "=" * ...` → `TypeError` (an empty string also hits `ValueError("input too short")`). Everywhere else in this function we use `.get(..., '')`. Should we guard for a missing/empty `pass` before revealing it here too?

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
DavidePrincipi: I think I got your point, but for the way the destination is structured and backup.webdav_write_file() is used, the immediate parent is enough. Iterative creation of the whole structure, like `mkdir -p` is not needed. We just need a straight `mkdir`-like behavior.

core/imageroot/usr/local/agent/pypkg/agent/__init__.py
DavidePrincipi: Yes it's intentional. I don't want to send secrets to logs in any case. I just want a DEBUG mode to inspect the restic command line, without the secrets, now.

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
DavidePrincipi: The newline is now a forbidden char in JSON schema validator

core/imageroot/usr/local/agent/pypkg/cluster/backup.py
DavidePrincipi: Now the code works on a podman_args copy.

stephdl (approved):

PR #1177, opened 2026-05-19
Enhance AddRepositoryModal for Rclone support and UI optimizations

stephdl:
Update AddRepositoryModal to support new storage providers, specifically Rclone, and improve UI translations and development build optimizations.

https://github.com/NethServer/dev/issues/7966


core/ui/public/i18n/en/translation.json
```diff
@@ -471,30 +471,31 @@
...
         "genericS3_short": "Generic S3",
-        "azure_short": "Azure",
         "download": "Download",
         "instances": "instance | instances",
         "Instances": "Instance | Instances",
         "applications": "Application | Applications",
-        "delete_backup_repository": "Delete backup destination",
+        "delete_destination": "Delete destination",
         "delete_repository_confirm": "Delete backup destination '{name}'?",
-        "delete_repo_explanation_1": "Any scheduled backup using this destination will be deleted too, but your backed up data will be left intact on the original destination. This action is NOT reversible.",
```
andre8244, line 485: Let's keep showing this warning message; the mockup is not showing it, but it is not intended to be removed.

core/ui/public/i18n/en/translation.json
```diff
@@ -471,30 +471,31 @@
...
-        "azure_url_label": "Azure container name",
+        "repository_name": "Destination name",
+        "repository_name_helper": "Name used to identify this destination",
         "aws_url_label": "AWS bucket name",
         "backblaze_url_label": "B2 bucket name",
         "genericS3_url_label": "Bucket address",
         "genericS3_url_tooltip": "Enter an address with format 'serverfqdn/bucketname'. Note: the server must have a valid TLS certificate",
         "genericS3_placeholder": "e.g. myminio.example.com/mybucket",
         "smb_url_label": "Share name",
+        "smb_protocol_warning_title": "Unsupported file share protocol",
```
andre8244: "SMB2 and SMB3 protocols supported"

core/ui/public/i18n/en/translation.json
```diff
@@ -660,6 +655,17 @@
...
+        "rclone_configuration_source_paste_configuration": "Paste configuration",
+        "rclone_configuration_content": "Rclone configuration",
+        "rclone_configuration_content_single_destination_helper_text": "Paste a configuration containing a single destination",
+        "rclone_configuration_source_upload_configuration": "Upload configuration file",
+        "rclone_configuration_file": "Rclone configuration file",
+        "rclone_configuration_file_helper": "Supported file types: .conf, .txt",
+        "rclone_conf_error": "Invalid rclone configuration",
+        "rclone_base_path": "Path",
+        "rclone_configuration_source_keep_configuration": "Keep current configuration",
+        "data_encryption_key_placeholder": "e.g a76917c4522...",
```
andre8244: Let's remove this string and use instead `$t('common.eg_value', { value: 'a76917c4522..."' })`

core/ui/src/components/backup/AddRepositoryModal.vue
```diff
@@ -433,6 +488,7 @@
                   "
                   v-model="password"
                   :helper-text="$t('backup.repository_password_helper')"
+                  :placeholder="$t('backup.data_encryption_key_placeholder')"
```
andre8244: Let's use `$t('common.eg_value', { value: 'a76917c4522..."' })` instead

core/ui/src/components/backup/AddRepositoryModal.vue
```diff
@@ -560,10 +616,13 @@ export default {
...
-        azure: {
-          azure_account_key: "",
-          azure_account_name: "",
+        rclone: {
+          rclone_conf_secret: "",
+          base_path: "",
+          configuration_source: "",
+          configuration_file: "",
         },
         //// handle all providers
```
andre8244: I think it is time to remove these comments (that I introduced a long time ago 😅)

core/ui/src/components/backup/AddRepositoryModal.vue
```diff
@@ -1464,4 +1608,8 @@ export default {
   width: 100%;
   height: 100%;
 }
+
+.maxwidth {
+  max-width: 38rem;
```
andre8244: It would be better to add `.cv-text-area` to [this CSS rule](https://github.com/NethServer/ns8-core/blob/main/core/ui/src/styles/_core.scss#L578-L587)

core/ui/src/components/backup/EditRepositoryModal.vue
```diff
@@ -532,14 +625,41 @@ export default {
...
+      this.error.rclone.configuration_file = "";
 
       let isValidationOk = true;
 
+      if (this.rclone.configuration_source === "paste") {
+        if (!this.rclone.rclone_conf_secret) {
+          this.error.rclone.rclone_conf_secret = this.$t("common.required");
+
+          if (isValidationOk) {
+            this.$nextTick(() => {
```
andre8244: Did you try if a simple `focusElement(rclone_conf_secret)` is enough?

core/ui/src/components/backup/RestoreSingleInstanceSnapshotSelector.vue
```diff
@@ -36,11 +36,23 @@
         @click="deselectOtherSnapshots(snapshot)"
         class="snapshot-tile"
       >
-        <div>
-          {{ formatSnapshotTimestamp(snapshot.timestamp) }}
+        <div class="snapshot-header">
```
andre8244: It's ok, but instead of defining new CSS classes (`snapshot-header`) we might use those defined in `core/ui/src/styles/_utils.scss`; here qui can use something like `class="flex items-center gap-2"`

andre8244 (approved):
- Thanks for cleaning up the code, we still have a couple of "//// handle all providers", try a global search on the project
- Not sure about the spinner showing after clicking "Download cluster backup". What do you think @AmaLuci ? Sadly the component does not currently support showing a spinner inside that button, shall we avoid showing it? The dowload preparation is usually fast, maybe we can avoid giving feedback to the user in this case

Image

AmaLuci:
> * Thanks for cleaning up the code, we still have a couple of "//// handle all providers", try a global search on the project
> * Not sure about the spinner showing after clicking "Download cluster backup". What do you think @AmaLuci ? Sadly the component does not currently support showing a spinner inside that button, shall we avoid showing it? The dowload preparation is usually fast, maybe we can avoid giving feedback to the user in this case
> 
> Image

I agree with you, it's not necessary. Let's hide the spinner.

PR #1189, opened 2026-05-27
feat: handle create-cluster network_check_failed

DavidePrincipi:
Validate network configuration before create-cluster alters Redis state. Show an UI notification error if network check fails.

Refs NethServer/dev#8021

### DIFF
--- a/core/ui/public/i18n/en/translation.json
+++ b/core/ui/public/i18n/en/translation.json
@@ -270,6 +270,7 @@
         "review_worker_fqdn_description": "Set the fully qualified domain name (FQDN) of this cluster worker",
         "not_a_private_network": "Must be included in the private IPv4 network addresses: 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16",
         "not_a_valid_ipv4_network": "Invalid IPv4 network address",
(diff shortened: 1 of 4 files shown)


DavidePrincipi:
image

core/ui/public/i18n/en/translation.json
```diff
@@ -270,6 +270,7 @@
         "review_worker_fqdn_description": "Set the fully qualified domain name (FQDN) of this cluster worker",
         "not_a_private_network": "Must be included in the private IPv4 network addresses: 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16",
         "not_a_valid_ipv4_network": "Invalid IPv4 network address",
+        "network_check_failed": "Internet connectivity check failed. Ensure the network is configured correctly and DNS resolution is working",
```
gsanchietti:
```suggestion
        "network_check_failed": "Network connectivity check failed. Ensure the node can access Internet and DNS resolution is working",
```

gsanchietti (approved):
I proposed a rewording, but not sure it fits better than the existing one ;)

andre8244 (commented):
What happens if the user clicks "Go back", then fixes the network connection, then goes back to this page? It looks like `error.networkCheckCreateCluster` is never reset.

andre8244 (approved):

PR #1191, opened 2026-05-27
Update permission checks for backup destination validation

stephdl:
Revise permission checks to enhance validation of backup destinations by changing the write access test to utilize a temporary file instead of a directory. This improves accuracy in permission validation.

https://github.com/NethServer/dev/issues/7854


DavidePrincipi (changes requested):
I'm sorry I cannot accept this behavior change.

Directory creation must be checked: it's the first thing we do on Rclone remotes, to build the two-level directory structure where Restic repo is placed.

In the future we can also check file creation if we notice that dir check is not enough and its assumption is weak.

PR #1192, opened 2026-05-28
Add loadingAction prop to NsInlineNotification component

stephdl:
Introduce a new `loadingAction` prop for the NsInlineNotification component and update story examples to demonstrate its usage.

https://github.com/NethServer/ns8-ui-lib/pull/78


andre8244 (approved):
Off-topic: since we are modifying NsInlineNotification storybook, we might set `lowContrast: true` in all stories, since that's the default value now

PR #1194, opened 2026-06-03
fix: work around systemd %S path change on Rocky 9.8

### LINKED ISSUE #8028: Systemd %S breaking changes in Rocky Linux 9.8
Newer release of Systemd RPM in Rocky Linux and Alma Linux 9.8 is not backward-compatible in the way it expands `%S`.

Expansion of %S changed from `%h/.config` to `%h/.local/state`, breaking many Systemd units.

Refs https://git.rockylinux.org/staging/src-rhel/rpms/systemd/-/blob/c9s/SOURCES/1313-execute-add-support-for-XDG_STATE_HOME-for-placing-s.patch?ref_type=heads

**Steps to reproduce**

- Update base distro to release 9.8:

       dnf --enablerepo=baseos,appstream --disablerepo=ns-baseos,ns-appstream update -y

- Restart a rootless application from UI, or reboot the node

**Expected behavior**

The rootless application is working as usual.

**Actual behavior**

The agent.service and application container restart fail.

```
Jun 03 10:08:45 rocky-linux98 systemd[13089]: Starting Rootless module/traefik1 agent...
Jun 03 10:08:45 rocky-linux98 systemd[13108]: agent.service: Failed at step CHDIR spawning /usr/bin/chmod: No such file or directory
Jun 03 10:08:45 rocky-linux98 systemd[13109]: agent.service: Failed at step CHDIR spawning /usr/local/bin/agent: No such file or directory
Jun 03 10:08:45 rocky-linux98 systemd[13089]: agent.service: Main process exited, code=exited, status=200/CHDIR
Jun 03 10:08:45 rocky-linux98 systemd[13089]: agent.service: Failed with result 'exit-code'.
```

**Components**

- Systemd systemd-252-67.el9_8.2.rocky.0.1.x86_64

**See also**

- Bug thread https://community.nethserver.org/t/upgrade-almalinux-9-7-to-9-8-didnt-work/27720
- Forum fix release announcement https://community.nethserver.org/t/ns8-core-3-19-1-released-required-compatibility-update-for-rocky-linux-and-almalinux-9-8/27773

----

Thanks to **CoBlau**

DavidePrincipi:
Rocky Linux 9.8 ships a systemd version where the `%S` specifier now expands to `~/.local/state` instead of `~/.config`, breaking rootless module agent and container startup due to path mismatch.

Replace `%S` with `%E` (`XDG_CONFIG_HOME`, `~/.config`) in agent.service and api-moduled.service. Add `fix-xdg-state@.service`, a one-shot unit that creates a `~/.local/state` -> `../.config` symlink before `user@.service` starts, preserving compatibility with any path already hardcoded elsewhere.

Remove the superseded `01-xdg-state.conf` `ManagerEnvironment=` override, which is no longer needed.

The symlink is created on the next user session restart, (i.e. next node reboot, or forced application restart). Until that happens the running Systemd session manager keeps the previous %S assignment and the bug is not present.

The symlink is also created when a new module is added to the system (systemd user session starts early), covering all possible creation paths.

Fixes NethServer/dev#8028

### DIFF
--- a/core/imageroot/etc/systemd/system/fix-xdg-state@.service
+++ b/core/imageroot/etc/systemd/system/fix-xdg-state@.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Create User-State-Private link
+Documentation=https://github.com/NethServer/dev/issues/8028
+
+[Service]
+User=%i
+Type=oneshot
+ExecStart=bash -c '! [[ -e ~/.local/state ]] && ln -sfv ../.config ~/.local/state'
+RemainAfterExit=true
--- a/core/imageroot/etc/systemd/system/fix-xdg-state@.service
+++ b/core/imageroot/etc/systemd/system/fix-xdg-state@.service
@@ -8,4 +8,3 @@ Type=oneshot
 WorkingDirectory=~
 ExecStart=bash -c '! [[ -d .local ]] && mkdir -vp .local'
 ExecStart=bash -c '! [[ -e .local/state ]] && ln -sfv ../.config .local/state'
-RemainAfterExit=true
--- a/core/imageroot/etc/systemd/system/fix-xdg-state@.service
+++ b/core/imageroot/etc/systemd/system/fix-xdg-state@.service
@@ -5,5 +5,7 @@ Documentation=https://github.com/NethServer/dev/issues/8028
 [Service]
 User=%i
 Type=oneshot
-ExecStart=bash -c '! [[ -e ~/.local/state ]] && ln -sfv ../.config ~/.local/state'
+WorkingDirectory=~
+ExecStart=bash -c '! [[ -d .local ]] && mkdir -vp .local'
+ExecStart=bash -c '! [[ -e .local/state ]] && ln -sfv ../.config .local/state'
 RemainAfterExit=true
--- a/core/imageroot/etc/systemd/system/fix-xdg-state@.service
+++ b/core/imageroot/etc/systemd/system/fix-xdg-state@.service
@@ -6,5 +6,5 @@ Documentation=https://github.com/NethServer/dev/issues/8028
 User=%i
 Type=oneshot
 WorkingDirectory=~
-ExecStart=bash -c '! [[ -d .local ]] && mkdir -vp .local'
-ExecStart=bash -c '! [[ -e .local/state ]] && ln -sfv ../.config .local/state'
+ExecStart=bash -c 'mkdir -vp .local'
+ExecStart=bash -c '[[ -e .local/state ]] || ln -sfv ../.config .local/state'
--- a/core/imageroot/etc/systemd/user.conf.d/01-xdg-state.conf
+++ b/core/imageroot/etc/systemd/user.conf.d/01-xdg-state.conf
@@ -1,3 +1,0 @@
-# This setting is required by Debian 13 and may raise a harmless warning in other distros.
-[Manager]
-ManagerEnvironment=XDG_STATE_HOME=%h/.config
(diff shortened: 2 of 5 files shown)


core/imageroot/etc/systemd/system/fix-xdg-state@.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Create User-State-Private link
+Documentation=https://github.com/NethServer/dev/issues/8028
+
+[Service]
+User=%i
+Type=oneshot
+WorkingDirectory=~
+ExecStart=bash -c '! [[ -d .local ]] && mkdir -vp .local'
+ExecStart=bash -c '! [[ -e .local/state ]] && ln -sfv ../.config .local/state'
```
stephdl: I wonder if the `! [[ ... ]] && cmd` pattern might silently fail on existing systems?

When `.local` already exists, `! [[ -d .local ]]` evaluates to exit code 1, and `&&` short-circuits with that same exit code 1 — so systemd considers the `ExecStart` as failed, without ever trying the second line.

Wouldn't something like this be safer?

```ini
ExecStart=bash -c 'mkdir -vp .local'
ExecStart=bash -c '[[ -e .local/state ]] || ln -sfv ../.config .local/state'
```

(`mkdir -p` is already idempotent and won't fail if `.local` exists.)

core/imageroot/etc/systemd/system/fix-xdg-state@.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Create User-State-Private link
+Documentation=https://github.com/NethServer/dev/issues/8028
+
+[Service]
+User=%i
+Type=oneshot
+WorkingDirectory=~
+ExecStart=bash -c '! [[ -d .local ]] && mkdir -vp .local'
```
stephdl: Should we use the full path `/bin/bash` here instead of relying on `PATH` resolution? Systemd's `PATH` in a system unit context is limited, and while `bash` should be findable, explicit paths feel more defensive for a unit that runs at session startup.

core/imageroot/etc/systemd/system/fix-xdg-state@.service
stephdl: What happens if `~/.local/state` already exists as a real directory (not a symlink)? For example on a system that had actual XDG state data written there by another application. The condition `[[ -e .local/state ]]` would be true, so the symlink would not be created and the system would remain broken — without any warning in the journal. Is that an expected edge case we can ignore, or should we at least log something?

core/imageroot/etc/systemd/user.conf.d/01-xdg-state.conf
```diff
@@ -1,3 +0,0 @@
-# This setting is required by Debian 13 and may raise a harmless warning in other distros.
```
stephdl, line 1: The comment in this file said it was *required* by Debian 13. Is the symlink approach sufficient to cover Debian 13 as well, or do we need to verify that separately before removing this?

stephdl (commented):
Nice approach overall — using `%E` + a compatibility symlink is clean. Just a few questions below.

core/imageroot/etc/systemd/system/fix-xdg-state@.service
gsanchietti: I agree with Steph, we can make it a little more safer?
```suggestion
ExecStart=bash -c 'mkdir -vp .local'
ExecStart=bash -c '! [[ -L .local/state ]] && ln -sfv ../.config .local/state'
```

DavidePrincipi: Not a real scenario: ~/.local/state does not exist in NS8

core/imageroot/etc/systemd/user.conf.d/01-xdg-state.conf
DavidePrincipi: Hopefully yes, let me test it

core/imageroot/etc/systemd/system/fix-xdg-state@.service
DavidePrincipi: existence check is enough: ln fails if state is a directory

gsanchietti (approved):

PR #1199, opened 2026-06-09
feat: run update-core on cluster init and restore

### LINKED ISSUE #8052: Prevent Core Apps version mismatch
The prebuilt image installation may start with an outdated NS8 core image. During `create-cluster`, the latest versions of Loki and Metrics are installed automatically, and their versions may be incompatible with the core version.

**Proposed solution**

Work around the issue by running `update-core` early during `create-cluster`, forcing the core components to be updated and aligned to the latest version before installing Loki and Metrics.

**Alternative solution**

Fail fast. Detect version mismatches and refuse to proceed. The installation must start from an up-to-date image.

**See also**

Discussion https://mattermost.nethesis.it/nethesis/pl/jyasi88p1pnrxqauqeejr3m8ko (splits NethServer/dev#8021)

DavidePrincipi:
When NS8 is installed from a prebuilt image, core and Traefik may be outdated. Add step 40update_cluster to create-cluster to force an update-core run immediately after cluster initialisation.

This also covers the disaster-recovery path: restore-cluster already calls create-cluster as a sub-task, so the new step is reached in both scenarios without duplication.

During core update the cluster agent receives SIGUSR1 and exits gracefully once all running actions complete, then systemd restarts it at the latest version.

Refs ~~NethServer/dev#8021~~ NethServer/dev#8052


gsanchietti (approved):

core/imageroot/var/lib/nethserver/cluster/actions/create-cluster/40update_cluster
```diff
@@ -0,0 +1,22 @@
...
+
+update_core_response = agent.tasks.run(
+    agent_id='cluster',
+    action='update-core',
+    data={'nodes': [1]}, # always 1 in a new cluster
+    endpoint="redis://cluster-leader",
+)
+if update_core_response['exit_code'] != 0:
+    print(agent.SD_ERR + "update-core failed", file=sys.stderr)
+    sys.exit(1)
```
gsanchietti, line 20: I'd prefer to not abort the create-cluster if the update fails.
Also make sure that is safe enough to call the update-core even if the cluster is not still completely configured.

andre8244 (approved):
UI part LGTM

DavidePrincipi:
I cannot merge this pull request: in my early tests the cluster-admin task progress (create-cluster, restore-cluster) hangs forever or logs out the user. I put the PR to draft, to investigate these behaviors further in a separate issue.

PR #1204, opened 2026-06-11
Add regular expression support for log filtering

stephdl:
Introduce regular expression support for log filtering and enhance UI feedback for log display. This update improves user experience by providing clearer feedback on the logs being shown.


stephdl:
wont fix

PR #1212, opened 2026-06-19
fix: do not request date-fns-locale packages on load

andre8244:
Fixed issue where many hundreds of date-fns-locale-* packages were requested on page load for all languages supported by date-fns

Refs:
- https://github.com/NethServer/dev/issues/8051
- Private conversation with @DavidePrincipi: https://mattermost.nethesis.it/nethesis/pl/sdznkezz1ff83npja9i8f65e7c

### DIFF
--- a/core/ui/src/views/settings/SettingsSubscription.vue
+++ b/core/ui/src/views/settings/SettingsSubscription.vue
@@ -351,8 +351,6 @@ import {
   PageTitleService,
   DateTimeService,
 } from "@nethserver/ns8-ui-lib";
-import { intervalToDuration, parseISO, formatDuration } from "date-fns";
-import { getDateFnsLocale } from "@/i18n";
 import { mapGetters } from "vuex";
 
 import NotificationService from "@/mixins/notification";
@@ -397,7 +395,6 @@ export default {
       termsUrl: "",
       agreeTerms: false,
       sessionExpireDate: "",
-      dateFnsLocale: null,
       loading: {
         getSubscription: false,
         setSubscription: false,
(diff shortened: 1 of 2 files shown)


core/ui/src/views/settings/SettingsSubscription.vue
```diff
@@ -439,24 +431,25 @@ export default {
...
         year: "numeric",
-      }).format(parsedDate);
-      const durationStr = formatDuration(duration, {
-        format: ["days", "hours"],
-        delimiter: ", ",
-        locale: this.dateFnsLocale,
-      });
+      }).format(end);
+
+      const durationStr = new Intl.DurationFormat(navigator.language, {
```
stephdl, line 449: `Intl.DurationFormat` is only available from Chrome 129 (Aug 2024), Firefox 136 (Feb 2025), and Safari 16.4 (Mar 2023) — roughly 84% global coverage according to caniuse. On an older browser this line throws a `TypeError` and breaks the subscription expiry display entirely.

Would it be worth adding a small guard?

```js
const durationStr =
  typeof Intl.DurationFormat !== "undefined"
    ? new Intl.DurationFormat(navigator.language, { style: "long" }).format({ days, hours })
    : `${days}d ${hours}h`;
```

`typeof` never throws even if the property doesn't exist, so the fallback (`"2d 5h"`) kicks in silently for unsupported browsers without crashing the component. Not a blocker for an admin UI, just a thought.

stephdl (commented):
Good fix overall — removing the dynamic `date-fns` locale import eliminates hundreds of spurious network requests on page load. The switch to native `Intl` APIs is the right direction.

core/ui/src/views/settings/SettingsSubscription.vue
andre8244: Good catch, but I truly hope that our users have updated their browser in the last 1-2 years... Otherwise they have much bigger issues :sweat_smile:

PR #1217, opened 2026-06-25
fix: restart systemd user sessions

DavidePrincipi:
Stop then start all non-root systemd user sessions during the update-core action.

This recreates /run/user/* and works around the SELinux labeling mismatch affecting netns directories after the Rocky Linux 9.8 upgrade.

:information_source:  This temporary workaround restarts systemd user sessions during the update and will be removed in a future core release. Implementation modifies an old script file to ease future removal of this code.


gsanchietti (approved):
🤞

Amygos (approved):

PR #1244, opened 2026-07-23
chore(deps): move Go toolchain to 1.26.5

gsanchietti:
## Description

Move the whole repo's Go toolchain to **1.26.5** (current stable) across all
Go components: builder, swagdoc, api-server and api-server-logs, agent,
api-moduled.

Also:
- integrate all pending renovate security updates
- bump golang dependencies to latest (gin-gonic, etc.)
- apply `go fix` to bring the code in line with the new toolchain's
  conventions, apply `go fmt`, cleanup with `mod tidy`

Most of this is mechanical, but a few real (small) behavior changes rode
along with it, worth calling out for reviewers:

- `go fix` added a `defer cancelBrpop()` in `listenActionsAsync`
  (`core/agent/htask.go`). This fixes a context/goroutine leak on agent
  shutdown — previously `cancelBrpop` was only invoked from the
  `actionsCtx.Done()` branch, so the path where `taskCh` closes never
  released the BRPOP context.
- `go fix` also turned several `log.Printf(SD_ERR + errorMessage)` calls into
  `log.Print(SD_ERR + errorMessage)` (`core/agent/hbuiltin.go`,
  `core/agent/htask.go`). The old code passed a dynamic, unsanitized string
  as a `Printf` format with no arguments, so a stray `%` in the message could
  mangle the log line. This changes log output for any message that
  contains `%`.
- `github.com/gin-contrib/static` bumped `v0.0.1` → `v1.1.6` (major version)
  in `core/api-server/go.mod`. The call site in `api-server.go`
  (`static.Serve("/", static.LocalFile(...))`) is untouched. Verified on a
  live cluster node (`curl -s http://127.0.0.1:9311/css/core.css` and `/`)
  that static/SPA asset serving still works as expected after the bump.


DavidePrincipi (commented):
Nice cleanup overall — the bulk of this is mechanical (`go fix`, `gofmt -s`, toolchain bump) and looks safe.

One ask: could you update the PR description to call out a few small-but-real behavior changes that rode along with the modernization, so reviewers/future readers don't mistake them for pure formatting?

1. **`core/agent/htask.go`** — `listenActionsAsync` now has a `defer cancelBrpop()` that wasn't there before. This actually fixes a context/goroutine leak on shutdown (previously `cancelBrpop` was only called from the `actionsCtx.Done()` branch), but it's new logic, not a reformat.
2. **`core/agent/hbuiltin.go` / `core/agent/htask.go`** — a few `log.Printf(SD_ERR + errorMessage)` calls became `log.Print(SD_ERR + errorMessage)`. Good fix (the old code passed a dynamic, unsanitized string as a `Printf` format with no args, so a stray `%` in the message could mangle the log line), but it does change log output for any message containing `%`.
3. **`core/api-server/go.mod`** — `github.com/gin-contrib/static` jumps from `v0.0.1` to `v1.1.6` (major version change), while the call site in `api-server.go` (`static.Serve("/", static.LocalFile(...))`) is untouched. Worth a quick manual check that SPA fallback / static serving behavior is unchanged, since that library's internals changed significantly across those versions.

None of these are blockers, just want the description to reflect that this isn't purely mechanical.

DavidePrincipi (approved):

PR #1245, opened 2026-07-23
feat(core): add update-core CLI command

gsanchietti:
## Summary
Adds an `update-core` CLI wrapper (`core/imageroot/usr/local/sbin/update-core`), mirroring the existing `update-module` command. Lets an admin push a specific core image to chosen nodes, or all nodes with `--all-nodes`.

## How to test
`update-core <core_url> <node_id> [<node_id> ...]` or `update-core <core_url> --all-nodes [--force] [--verbose]` on a cluster leader.

### DIFF
--- a/core/imageroot/usr/local/sbin/update-core
+++ b/core/imageroot/usr/local/sbin/update-core
@@ -1,0 +1,49 @@
+#!/usr/local/bin/runagent python3
+
+#
+# Copyright (C) 2026 Nethesis S.r.l.
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import sys
+import agent
+import agent.tasks
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('core_url')
+parser.add_argument('--nodes', nargs='+', type=int, help='update only these node identifiers, instead of every node in the cluster (edge case, use with care: nodes end up on different core versions)')
+parser.add_argument('--force', help='force update even if core_url is already present locally', action='store_true')
+parser.add_argument('--verbose', help='show update-core action stderr on success', action='store_true')
+args = parser.parse_args()
+
+image_name = agent.get_image_name_from_url(args.core_url)
+if image_name != 'core':
+    parser.error(f'core_url must reference the "core" image, got "{image_name}"')
+
+rdb = agent.redis_connect()
+if args.nodes:
+    nodes = args.nodes
+else:
+    nodes = list(int(node_id) for node_id in set(rdb.hvals('cluster/module_node')))
+
+result = agent.tasks.run(
+    agent_id='cluster',
+    action='update-core',
+    data={
+        'core_url': args.core_url,
+        'nodes': nodes,
+        'force': args.force,
+    },
+    extra={
+        'title': "cluster/update-core",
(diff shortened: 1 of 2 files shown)


core/imageroot/usr/local/sbin/update-core
```diff
@@ -0,0 +1,51 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import sys
+import agent
+import agent.tasks
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('core_url', nargs='?', default='', help='core image URL to install; if omitted, the latest core image from the repository is used')
```
stephdl: With `core_url` now optional (`nargs='?'`) sitting right before the variadic `nodes` (`nargs='*'`), argparse can't tell them apart when `core_url` is omitted together with a node list.

`update-core 2 4` — my intent being "update nodes 2 and 4 with the latest core" — parses as `core_url='2'`, `nodes=[4]`. It ends up acting on node 4 only (with a bogus image ref `'2'`) and silently skips node 2. The omit-`core_url` path only works with `--all-nodes`; omit + explicit node list is unusable.

Would turning `core_url` into an option rather than a positional solve it? Something like `update-core --core-url ghcr.io/... 2 4`, so `update-core 2 4` unambiguously means "latest core on nodes 2 and 4". WDYT?

gsanchietti: We could to it, but then it will be something like:
```
update-core --core-url ghcr.io/nethserver/core:MAJ.MIN.PATCH --all-nodes
```

Which is different from `update-module` behavior.
What about removing completely the automatic mode, and require always the core URL as argument (just like update-module)?

stephdl: What about keeping `core_url` as a required positional (exactly like `update-module`) but keeping `--all-nodes` on top?

```
update-core ghcr.io/nethserver/core:3.20.1 2 4          # pinned core, nodes 2 and 4
update-core ghcr.io/nethserver/core:3.20.1 --all-nodes  # pinned core, whole cluster
```

Required `core_url` consumes the first token, `nodes` takes the rest — no argparse ambiguity, and the `2 4` footgun is gone. We drop the auto/latest mode, but we stay consistent with `update-module` and gain the fleet-wide `--all-nodes` convenience. It's also safer: the operator always states the exact image explicitly, no surprise auto-latest.

gsanchietti: Done: reverted 2 last commits

stephdl (approved):
great work

DavidePrincipi (commented):
The expected outcome of update-core is that all nodes are aligned at the same core version. Cherry-picking some nodes and leaving back others is a really development edge case. Hence I'd remove `--all-nodes` flag making its behavior the default one. And add `--nodes` optional flag with a list of node IDs to cover the edge case.

Also please add a check for the image name, that must be "core", just in case somebody wants to install Mattermost as core image.

Note that the core_url action attribute is optional. If missing it is automatically resolved to "the latest available core version": I'd keep this behavior in the helper too.

DavidePrincipi (approved):

PR #1247, opened 2026-07-24
feat(software-center): restructure updates tab and expose automatic updates opt-out

stephdl:
Implement the Software Center UI for automatic updates, per the approved design in NethServer/dev#7976. Frontend only; the backend policy is already enforced on the base branch `feat-8101`.

## Changes

- **Updates tab**: two stacked sections, **Core update** and **Applications update**, replace the banner pointing at a separate page. Each renders only when it has something to show, with an explicit empty state.
- **Cluster-wide automatic updates**: enable and disable from the page kebab menu, with a confirmation modal in both directions and a warning notification while they are off.
- **Per-application automatic updates**: same modal from the **Applications** and per-app instances views, with an "Updates disabled" tag next to the version. Tag and menu entry read a single gate (active subscription plus cluster-wide updates on), since the per-application opt-out has no effect while the timer is off.
- **Notifications**: one notification covers core and application updates. Previously the core one hid the application one.
- **Failure paths**: every terminal outcome of `set-automatic-updates` releases the modal; an aborted `list-modules` is reported instead of leaving the flags silently false.
- **Other**: page title tooltip removed (not in the design), pause and play icons on the toggle entry, no more `NavigationDuplicated` when clicking the active side menu entry.

New file: `core/ui/src/components/software-center/AutomaticUpdatesModal.vue` — one modal for both scopes; an `instance` prop switches payload and wording.

## Backend contract consumed
- `list-modules`: `apply_updates_is_active`, `subscription_is_active`, `installed[].automatic_updates`.
- `set-automatic-updates`: `{ apply_updates_is_active?, instances?: { id: bool } }`.

Refs: NethServer/dev#8101
Design: NethServer/dev#7976


core/ui/src/components/software-center/AutomaticUpdatesModal.vue
```diff
@@ -0,0 +1,210 @@
...
+      );
+      this.registerTaskEvent(
+        `${taskAction}-validation-failed-${eventId}`,
+        (validationErrors) => {
+          console.error(`${taskAction} validation failed`, validationErrors);
+          this.onTaskEnded();
+          this.error.setAutomaticUpdates = this.$t("error.generic_error");
+        }
+      );
+      this.taskTimeout = setTimeout(() => {
```
andre8244: This task management looks a bit overcomplicated, is there a real benefit in using it? Can we handle it like all other tasks, e.g. without `TASK_TIMEOUT` and `setTimeout`?

core/ui/src/styles/_core.scss
```diff
@@ -302,6 +302,13 @@ p {
   max-width: none;
 }
 
+// the inline svg sits on the text baseline, so the descender gap pushes the icon
+// above the label: make the wrapper a flex box so it is exactly the svg height
+.bx--overflow-menu-options .ns-menu-item .icon {
```
andre8244: Is this needed? Do you have a comparison screenshot with and without the rule?

core/ui/src/views/ApplicationsCenter.vue
```diff
@@ -238,13 +238,21 @@
...
                   </cv-data-table-cell>
                   <cv-data-table-cell>
-                    <span>{{ row.version }}</span>
-                    <cv-tag
-                      v-if="row.update"
-                      class="mg-left-sm"
-                      kind="green"
-                      :label="$t('applications.update_available')"
-                    />
+                    <div class="version-cell">
```
andre8244: We can use `class="flex items-center"` and remove `.version-cell' rule

core/ui/src/views/SoftwareCenter.vue
```diff
@@ -282,29 +261,68 @@
...
+                :apps="[coreApp]"
+                :skeleton="false"
+                tab="updates"
+                key="core-update"
+                :light="true"
+              />
+            </template>
+            <!-- apps updates -->
+            <template v-if="appUpdates.length && !loadingUpdates">
+              <h4 class="mg-bottom-sm">
```
andre8244:
```suggestion
              <h4 class="mg-bottom-lg">
```

core/ui/src/views/SoftwareCenter.vue
```diff
@@ -282,29 +261,68 @@
...
-            >
-              <NsButton
-                kind="primary"
-                :icon="Upgrade20"
-                @click="willUpdateAllApps()"
-                :disabled="isUpdateInProgress"
-                >{{ $t("software_center.update_all_apps") }}</NsButton
+            <!-- core update -->
+            <template v-if="isCoreUpdateAvailable && !loadingUpdates">
+              <h4 class="mg-bottom-sm">
```
andre8244:
```suggestion
              <h4 class="mg-bottom-lg">
```

core/ui/src/views/SoftwareCenter.vue
```diff
@@ -282,29 +261,68 @@
...
+                key="core-update"
+                :light="true"
+              />
+            </template>
+            <!-- apps updates -->
+            <template v-if="appUpdates.length && !loadingUpdates">
+              <h4 class="mg-bottom-sm">
+                {{ $t("software_center.applications_update") }}
+              </h4>
+              <p
```
andre8244, line 285: According to the design this description should be gray (`$text-02`)

core/ui/src/views/SoftwareCenter.vue
```diff
@@ -282,29 +261,68 @@
...
-                :icon="Upgrade20"
-                @click="willUpdateAllApps()"
-                :disabled="isUpdateInProgress"
-                >{{ $t("software_center.update_all_apps") }}</NsButton
+            <!-- core update -->
+            <template v-if="isCoreUpdateAvailable && !loadingUpdates">
+              <h4 class="mg-bottom-sm">
+                {{ $t("software_center.core_update") }}
+              </h4>
+              <p class="mg-bottom-lg">
```
andre8244: According to the design this description should be gray (`$text-02`)

core/ui/src/views/SoftwareCenterAppInstances.vue
```diff
@@ -318,8 +336,16 @@
                       >{{ $t("common.node") }} {{ instance.node }}</span
                     >
                   </div>
-                  <div class="row">
-                    {{ $t("common.version") }} {{ instance.version }}
+                  <div class="row version-row">
+                    <span
+                      >{{ $t("common.version") }} {{ instance.version }}</span
+                    >
+                    <cv-tag
```
andre8244: Let's move this tag above the "Release notes" button (we already have @AmaLuci's blessing)

core/ui/src/components/software-center/AutomaticUpdatesModal.vue
stephdl: Simplified in 1efb4c9d: no more TASK_TIMEOUT, setTimeout or event registry, just the standard $root.$once completed/aborted pattern like SetInstanceLabelModal. I kept the validation-failed listener, otherwise a rejected input leaves the modal spinning forever. Tell me if you want it gone too.

core/ui/src/views/ApplicationsCenter.vue
stephdl: Done in cd44182e: class="flex items-center", .version-cell rule removed.

core/ui/src/views/SoftwareCenter.vue
stephdl: Applied in c8e29ff6.

core/ui/src/views/SoftwareCenter.vue
stephdl: Applied in c8e29ff6.

core/ui/src/views/SoftwareCenter.vue
stephdl: Done in c8e29ff6, via the existing .title-description class ($text-02).

core/ui/src/views/SoftwareCenter.vue
stephdl: Done in c8e29ff6, via the existing .title-description class ($text-02).

core/ui/src/views/SoftwareCenterAppInstances.vue
stephdl: Done in 031d4b9d: the tag now has its own row, right above the Release notes button. The .version-row rule became useless and is removed.

core/ui/src/styles/_core.scss
stephdl: Measured it on a live node instead of guessing. .ns-menu-item is already display:flex/align-items:center, but the .icon wrapper is display:block with line-height 18px, so its box is 24px tall around a 20px svg and the icon lands 2px above the label centre (icon centre 550.5 vs label centre 552.5). With the rule the wrapper is exactly 20px and the delta is 0. So the rule is doing something real, just small. I have 3x before/after crops of the kebab menu, stephdl will attach them here since images cannot be uploaded through the API.

stephdl: Follow-up: I compared the before/after crops and the 2px shift is not perceptible, so the rule was not worth a Carbon selector in _core.scss. Reverted in 666a43b3. If we ever want it, the .icon wrapper belongs to NsMenuItem in ns8-ui-lib, so the fix should live there and cover every menu at once.

andre8244 (approved):

PR #1257, opened 2026-07-28
TLS certificates page: ACME settings tab with challenge type selection

stephdl:
## Description

The TLS certificates page had a separate `ACME servers` page reachable through a top-right button, and the ACME challenge type — already accepted by `set-acme-server` — was exposed nowhere, so `TLS-ALPN-01` (needed when port 80 is unavailable) required editing `traefik.yaml` by hand.

This PR merges the two pages into two tabs, `Certificates` and `ACME settings`, and exposes the challenge type as a tag per node plus a required radio group in the edit modal. Options come from a single `ACME_CHALLENGE_TYPES` array, so `DNS-01` is one entry plus one translation key away; it is left out here because it needs extra fields. The ACME table columns become `Node`, `ACME directory URL`, `Challenge type`, which also fixes the previous wrong order, and `/settings/acme-servers` redirects to `/settings/tls-certificates?view=acme`.

Settings remain per node. No backend change: `set-acme-server` already validates `challenge` and `get-acme-server` already returns it.

## Surviving the Traefik restart

`set-acme-server` restarts Traefik, which serves the admin UI and its websocket, so the completion event carrying the table refresh is lost: stale values and skeleton rows until a manual reload. The Certificates tab has the same problem through `cert_helpers.purge_acme_json_and_restart_traefik`. Both tabs now watch the Vuex `isWebsocketConnected` flag and reload on reconnection; counters are clamped and ACME rows keyed by Traefik instance, so a late event cannot unbalance a counter or duplicate a row. The `Reload required` warning is gone, and the restart warning is no longer gated on the leader node — the restart happens on whichever node is edited. Mechanism borrowed from #1258, which fixes the same thing on HTTP routes; a shared mixin is the follow-up once both land.

## Design

Mockups: [Figma](https://www.figma.com/design/OsYqPXdkKJz0tBJRsiQgAf/NethServer-shared-file?node-id=4575-55402), design issue NethServer/dev#7973.

Refs: NethServer/dev#8099


core/ui/public/i18n/en/translation.json
```diff
@@ -931,16 +933,20 @@
         "certificates_not_displayed": "Certificates configured on {instanceId} are not displayed"
     },
     "settings_acme_servers": {
-        "title": "ACME servers",
-        "search_acme_server": "Search ACME server",
-        "no_acme_server": "No ACME server",
-        "no_acme_server_description": "There is no ACME server configured",
-        "url": "URL",
+        "acme_settings": "ACME settings",
+        "search_acme_settings": "Search ACME settings",
```
```suggestion
        "search_acme_settings": "Search",
```

core/ui/public/i18n/en/translation.json
```diff
@@ -931,16 +933,20 @@
...
-        "edit_acme_server": "Edit ACME server",
-        "edit_acme_server_for_node": "Edit ACME server for {node}",
-        "reload_page": "Page reload needed",
-        "edit_acme_server_leader_node_warning": "You are editing the ACME server for the leader node. After you click 'Edit ACME server' you'll need to reload the page"
+        "challenge": "Challenge type",
+        "challenge_http_01": "HTTP-01",
+        "challenge_tls_alpn_01": "TLS-ALPN-01",
+        "edit_acme_settings": "Edit ACME settings",
+        "acme_settings_for_node": "ACME settings for {node}",
+        "acme_restart_message": "HTTP clients on {node} will be briefly disconnected. Reload the page if necessary.",
```
```suggestion
        "acme_restart_message": "HTTP clients on {node} will be briefly disconnected. The page should reload automatically — otherwise, reload it manually..",
```

core/ui/src/components/settings/AcmeSettings.vue
```diff
@@ -0,0 +1,317 @@
...
+            <cv-data-table-cell>
+              <NsTag
+                v-if="row.challenge"
+                :kind="getChallengeTagKind(row.challenge)"
+                size="sm"
+                :label="getChallengeLabel(row.challenge)"
+              />
+            </cv-data-table-cell>
+            <cv-data-table-cell>
+              <div class="justify-flex-end">
```
andre8244: We can use `class="flex justify-end"`, these class already exist here https://github.com/NethServer/ns8-core/blob/main/core/ui/src/styles/_utils.scss

core/ui/src/components/settings/AcmeSettings.vue
```diff
@@ -0,0 +1,317 @@
...
+      this.getAcmeServer();
+    },
+  },
+};
+</script>
+
+<style scoped lang="scss">
+@import "../../styles/carbon-utils";
+
+.justify-flex-end {
```
andre8244: We can remove this class

core/ui/src/views/settings/SettingsTlsCertificates.vue
```diff
@@ -1375,4 +1518,11 @@ export default {
 .icon-and-text {
   justify-content: flex-start;
 }
+
+// carbon renders the interactive tooltip trigger as a button with a default
+// cursor, which hides the fact the tooltip opens on click
+.clickable-tooltip ::v-deep .bx--tooltip__trigger,
```
andre8244: Are these CSS rules needed? 🤔

core/ui/src/views/settings/SettingsTlsCertificates.vue
```diff
@@ -52,396 +40,488 @@
...
-                    <template v-else>
-                      <!-- no tls certificates -->
-                      <NsEmptyState
+          <NsTabs
+            :container="false"
+            :aria-label="$t('common.tab_navigation')"
+            :noDefaultToFirst="true"
+            @tab-selected="tabSelected"
+          >
+            <cv-tab
```
andre8244, line 49: This file has become very big (~1500 loc), maybe we can try to externalize two components, e.g. `TlsCertficatesPanel` and `AcmeSettingsPanel`

core/ui/public/i18n/en/translation.json
stephdl: Applied, minus the double period at the end.

One thing to check: the app never reloads by itself. When Traefik restarts on the leader, the websocket drops and `App.vue` shows a persistent warning toast with a "Reload" action, then retries the connection after 10 seconds. So the page recovers on its own, but it does not reload. Is "reload automatically" still the wording you want, or would "reconnect automatically" be closer?

core/ui/src/views/settings/SettingsTlsCertificates.vue
stephdl: Half of them were, yes. I measured it in the browser: carbon already sets `cursor: pointer` via `.bx--tooltip__trigger:not(.bx--btn--icon-only)`, so the rule on the trigger did nothing. I removed it.

But `.bx--tag` sets `cursor: default`, so the pointer was lost when hovering the "Obsolete" tag itself, which is the trigger content. I kept one narrow rule for that case:

```scss
// carbon's .bx--tag forces a default cursor, hiding that the tooltip opens on click
.clickable-tag ::v-deep .bx--tag {
  cursor: pointer;
}
```

Happy to drop it too if you would rather not special-case the tag.

core/ui/src/views/settings/SettingsTlsCertificates.vue
stephdl: Done. `SettingsTlsCertificates.vue` is down to 289 lines: page chrome, tab state, and the `list-installed-modules` chain whose Traefik instance list both tabs consume. The certificates tab moved to `TlsCertificatesPanel.vue`, and `AcmeSettings.vue` is now `AcmeSettingsPanel.vue` for symmetry.

Two notes on the wiring:

- The node filter is bound to a query param, and `watchQueryData` is bound to the view instance in `beforeRouteEnter`, so the view still owns `selectedNodeId` and passes it down with `.sync`.
- The panel now runs `listCertificates()` from a watcher on `traefikInstances` instead of being called at the end of `listInstalledModulesCompleted`, same trigger `AcmeSettingsPanel` already used.

I also dropped the `selectedNodeLabel` computed while moving the code; it was defined and never referenced, already dead on main.

Tested on a live single-node cluster: both tabs, filters, query params, the upload modal and its error mapping, the tooltips, and the ACME edit modal.

core/ui/public/i18n/en/translation.json
andre8244: Ok, but IIRC the websocket warning toast disappears as soon as the connection is resumed, right?

Anyway, for the message I propose "HTTP clients on {node} will be briefly disconnected. The connection should resume automatically — otherwise, reload the page manually."

core/ui/src/views/settings/SettingsTlsCertificates.vue
andre8244: IIRC the tooltip is shown on click, not on hover, right? So we need a visual cue (the cursor pointer) to show the element as interactive. Let's keep only the `.clickable-tag ::v-deep .bx--tag` rule.

core/ui/public/i18n/en/translation.json
stephdl: > Ok, but IIRC the websocket warning toast disappears as soon as the connection is resumed, right?

Yes and once reconnected we trigger all the list-*/get-* function

stephdl: Done, applied your wording.

andre8244 (approved):

PR #1258, opened 2026-07-28
Add Frontend proxies tab to the HTTP routes page

stephdl:
## Summary

Expose the Traefik trusted proxies settings in the cluster admin UI. Until now the only way to configure them was `api-cli run module/traefik1/set-trusted-proxies`.

The **HTTP routes** page gets a tab bar: the existing route table moves into a **Routes** tab, and a new **Frontend proxies** tab is added. The selected tab is kept in the `view` query parameter.

The new tab lists one row per node running behind a frontend HTTP proxy, with its proxy addresses and its trust depth, plus a text filter and a node filter. The Add/Edit modal takes the node, the proxy addresses (one per line) and the trust depth. The Add dropdown only offers nodes that are not configured yet, since `set-trusted-proxies` replaces the whole list of a node. Deleting a row sends an empty proxy list.

UI-only change: no Traefik action, schema or configuration template is touched.

## Related issue

NethServer/dev#8098


core/ui/src/views/settings/SettingsHttpRoutes.vue
```diff
@@ -59,196 +60,226 @@
...
-                              alignment="center"
-                              direction="top"
-                              class="shrink-0"
-                            >
-                              <template #trigger>
-                                <WarningAltFilled16 class="ns-warning" />
-                              </template>
-                              <template #content>
+                        <template slot="empty-state">
+                          <NsEmptyState
```
andre8244: Missing pictogram in the empty state

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +847,29 @@
...
+        "add_frontend_proxy_no_node_message": "No cluster node with a Traefik instance is available",
+        "edit_frontend_proxy_node": "Edit frontend proxies of {node}",
+        "delete_frontend_proxy_node": "Delete frontend proxies of {node}",
+        "delete_frontend_proxy_description": "Delete the frontend proxies of {node}? The real client IP won't be read from the X-Forwarded-For header anymore.",
+        "no_frontend_proxy": "No frontend proxy configured",
+        "no_frontend_proxy_description": "Add frontend proxy addresses to correctly identify client IPs behind reverse proxies or load balancers.",
+        "filter_frontend_proxies": "Filter",
+        "choose_node": "Choose node",
+        "frontend_proxies_placeholder": "e.g. 192.168.1.10\n2001:db8::1",
+        "frontend_proxies_helper": "Enter one IPv4 or IPv6 address per line",
```
andre8244, line 862: Are CIDR networks supported? They should be, according to the mockups

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +847,29 @@
...
+        "trust_depth": "Trust depth",
+        "trust_depth_helper": "Number of frontend proxy levels",
+        "trust_depth_tooltip": "Number of frontend proxies between the client and Traefik. With a trust depth of 1, the client IP address is read from the last entry of the X-Forwarded-For header.",
+        "trust_depth_invalid": "Enter a whole number greater than or equal to 0",
+        "bad_ip_address": "Invalid IP address. Enter one IPv4 or IPv6 address per line, without CIDR prefix",
+        "frontend_proxy_restart_message": "HTTP clients connected to {node} will be briefly disconnected. If your session goes through this node, the page reconnects on its own."
```
andre8244: IIUC, we decided to show "Reload the page if necessary" instead of "If your session goes through this node, the page reconnects on its own"

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
```diff
@@ -0,0 +1,410 @@
...
+        <cv-number-input
+          v-model="depth"
+          :label="''"
+          :aria-label="$t('settings_http_routes.trust_depth')"
+          :helper-text="$t('settings_http_routes.trust_depth_helper')"
+          :invalid-message="error.depth"
+          :disabled="loading.setTrustedProxies"
+          :min="0"
+          :step="1"
+          class="trust-depth mg-bottom-md"
```
andre8244: Carbon fields already provide a margin-bottom, which is larger than `mg-bottom-md`
```suggestion
          class="trust-depth"
```

core/ui/public/i18n/en/translation.json
stephdl: Not at the moment: `set-trusted-proxies` validates with `ipaddress.ip_address()`, which
rejects any `/prefix`. Traefik itself accepts CIDR in `trustedIPs`, so supporting it means
relaxing the check in ns8-traefik.

andre8244:
is `bad_ip_address` i18n string used (maybe by the backend)? If not, we can remove it

stephdl:
Yes, by the backend. `set-trusted-proxies` returns `error: bad_ip_address`, and
`setTrustedProxiesValidationFailed()` resolves the key dynamically via
`getI18nStringWithFallback()` — same pattern as the other modals, so grep finds nothing.
Removing it would show the raw key on screen.

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
```diff
@@ -0,0 +1,411 @@
...
+            :title="$t('action.set-trusted-proxies')"
+            :description="error.setTrustedProxies"
+            :showCloseButton="false"
+          />
+        </div>
+      </cv-form>
+    </template>
+    <template slot="secondary-button">{{ $t("common.cancel") }}</template>
+    <template slot="primary-button">{{
+      isEditing ? $t("common.save") : $t("common.add")
```
andre8244:
```suggestion
      isEditing ? $t("common.save") : $t("common.add_frontend_proxy")
```

core/ui/public/i18n/en/translation.json
```diff
@@ -18,6 +18,7 @@
         "clear_search": "Clear search",
         "settings": "Settings",
         "cancel": "Cancel",
+        "add": "Add",
```
```suggestion
```

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +847,30 @@
...
+        "add_frontend_proxy_no_node_message": "No cluster node with a Traefik instance is available",
+        "edit_frontend_proxy_node": "Edit frontend proxies of {node}",
+        "delete_frontend_proxy_node": "Delete frontend proxies of {node}",
+        "delete_frontend_proxy_description": "Delete the frontend proxies of {node}? Traefik will no longer see the real client IP address.",
+        "frontend_proxies_description": "A cluster node can stand behind a reverse proxy or a load balancer. Traefik trusts the X-Forwarded-For header of the addresses listed here to recover the original client IP.",
+        "no_frontend_proxy": "No frontend proxy configured",
+        "no_frontend_proxy_description": "Add the addresses of the proxies in front of a cluster node.",
+        "filter_frontend_proxies": "Filter",
+        "choose_node": "Choose node",
+        "frontend_proxies_placeholder": "e.g. 192.168.1.10\n2001:db8::1",
```
andre8244: Let's use `common.eg_value` instead

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +847,30 @@
...
+        "frontend_proxies_helper": "Enter one IPv4 or IPv6 address per line",
+        "trust_depth": "Trust depth",
+        "trust_depth_helper": "Number of frontend proxy levels",
+        "trust_depth_tooltip": "With a trust depth of 1, the client IP address is read from the last entry of the X-Forwarded-For header.",
+        "trust_depth_invalid": "Enter a whole number greater than or equal to 1",
```
```suggestion
        "trust_depth_invalid": "Enter a number of 1 or higher",
```

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
```diff
@@ -0,0 +1,411 @@
...
+</template>
+
+<script>
+import to from "await-to-js";
+import { UtilService, TaskService } from "@nethserver/ns8-ui-lib";
+import { mapState } from "vuex";
+
+// set-trusted-proxies validates with ipaddress.ip_address(): no CIDR ranges, and
+// no leading-zero octets, which Python rejects since 3.9.5
+const IPV4_OCTET = "(25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])";
```
andre8244: How about moving these constants to a `ip_address.js` lib or mixin? Also core/ui/src/components/settings/CreateOrEditHttpRouteModal.vue uses somethinh similar (`ipv4Pattern` and `ipv6Pattern` variables), maybe we can deduplicate the code.

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +847,30 @@
...
+        "bad_ip_address": "Invalid IP address. Enter one IPv4 or IPv6 address per line, without CIDR prefix",
+        "frontend_proxy_restart_message": "HTTP clients connected to {node} will be briefly disconnected. Reload the page if necessary."
```
```suggestion
        "frontend_proxy_restart_message": "HTTP clients connected to {node} will be briefly disconnected. The page should reload automatically — otherwise, reload it manually.."
```

core/ui/src/views/settings/SettingsHttpRoutes.vue
```diff
@@ -59,196 +60,229 @@
...
-                    <template slot="empty-state">
-                      <NsEmptyState
-                        :title="$t('settings_http_routes.no_http_route')"
+          <NsTabs
+            :container="false"
+            :aria-label="$t('common.tab_navigation')"
+            :noDefaultToFirst="true"
+            @tab-selected="tabSelected"
+          >
+            <cv-tab
```
andre8244, line 47: How about externalizing a `HttpRoutesPanel` component as well?

core/ui/src/components/settings/HttpFrontendProxiesPanel.vue
```diff
@@ -0,0 +1,684 @@
+<!--
```
andre8244, line 1: How about renaming to `HttpFrontendProxiesPanel`?

core/ui/src/views/settings/SettingsHttpRoutes.vue
stephdl: Fixed. The pictogram was there but node_modules was stuck on ns8-ui-lib 1.12.4, which has no NetworkPictogram. Works with 2.0.1.

core/ui/public/i18n/en/translation.json
stephdl: Not supported. set-trusted-proxies validates with ipaddress.ip_address(), which rejects any prefix. Supporting CIDR means relaxing that check in ns8-traefik. Out of scope here. The new mixin already has isIpAddressOrCidr, so the UI side is one line. I will open an issue on ns8-traefik.

core/ui/public/i18n/en/translation.json
stephdl: Superseded by your newer suggestion below. Applied that one.

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
stephdl: Done.

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
stephdl: Done. The key lives in common now, and common.add is gone.

core/ui/public/i18n/en/translation.json
stephdl: Done.

core/ui/public/i18n/en/translation.json
stephdl: Done. Placeholder is now common.eg_value with 192.168.1.10. Dropped frontend_proxies_placeholder. The IPv6 example is still in the helper text.

core/ui/public/i18n/en/translation.json
stephdl: Done.

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
stephdl: Done. New mixin in core/ui/src/mixins/ipAddress.js. There is no src/lib in this project, so I followed the mixins convention. It exposes isIpAddress (no prefix, what set-trusted-proxies accepts) and isIpAddressOrCidr, plus looksLikeIpv6 to pick the error message. Both modals use it now. One side effect: the route allow list rejects leading-zero octets like 192.168.001.1, which Python ipaddress refuses too.

core/ui/public/i18n/en/translation.json
stephdl: Done, with the trailing double period fixed to one.

core/ui/src/views/settings/SettingsHttpRoutes.vue
stephdl: Done. HttpRoutesPanel holds the tab body and its three modals. The view keeps the query params, the tab selector and the single list-installed-modules chain; each panel reloads from a traefikInstances watcher. The node filter is still a query param, so the panel takes it as a prop and mirrors it locally, because repainting the combo box needs a synchronous write.

core/ui/src/components/settings/HttpFrontendProxiesPanel.vue
stephdl: Done.

stephdl:
Pushed. Four commits.

Also reworked the routes tab to match the mockup and the frontend proxies tab: search box, node combo and a clear filters link on one row, no search box inside the data table, primary create button. Filtering to nothing shows a no results state. The node filter notification above the tabs is gone, the clear filters link replaces it.

Two things worth a look:
- The route allow list now rejects leading-zero octets such as `192.168.001.1`. Side effect of sharing one strict IPv4 pattern. Traefik would reject them anyway.
- The ns8-ui-lib bump to 2.0.1 carries the Vue 2.7 major. core/ui is already on `vue: ^2.7.0`, so nothing breaks.

Verified against a cluster with the dev server: both tabs, filters, query params after reload, tab switching, IP validation, CIDR still accepted in the route allow list.

core/ui/src/components/settings/ConfigureFrontendProxyModal.vue
andre8244: Sorry, my suggestion was not accurate, `add_frontend_proxy` belongs to `settings_http_routes` section, not `common`

core/ui/public/i18n/en/translation.json
```diff
@@ -844,7 +844,28 @@
...
+        "no_frontend_proxy_description": "Add the addresses of the proxies in front of a cluster node.",
+        "filter_frontend_proxies": "Filter",
+        "choose_node": "Choose node",
+        "frontend_proxies_helper": "Enter one IPv4 or IPv6 address per line",
+        "trust_depth": "Trust depth",
+        "trust_depth_helper": "Number of frontend proxy levels",
+        "trust_depth_tooltip": "With a trust depth of 1, the client IP address is read from the last entry of the X-Forwarded-For header.",
+        "trust_depth_invalid": "Enter a number of 1 or higher",
+        "bad_ip_address": "Invalid IP address. Enter one IPv4 or IPv6 address per line, without CIDR prefix",
+        "frontend_proxy_restart_message": "HTTP clients connected to {node} will be briefly disconnected. The page should reload automatically — otherwise, reload it manually."
```
andre8244: Let's fix this like in PR 1257

andre8244 (approved):
Let's fix a i18n string then LGTM

core/ui/public/i18n/en/translation.json
stephdl: done

PR #1265, opened 2026-07-31
fix: drop removed modules from backup schedules

DavidePrincipi:
## Summary
- Removed app instances used to keep lingering in a backup schedule's `instances` list and in per-node `backup_status` entries, so `list-backups` (and the UI) kept reporting them as backed up after the instance was uninstalled.
- Purge the module id from `cluster/backup/*` and `backup_status` when a module or a whole node is removed, and skip any still-stale reference in `list-backups` by checking `cluster/module_uuid`.
- Add a one-time-per-install, idempotent migration that prunes pre-existing dangling references (from before this fix) using `cluster/module_node` as the authoritative source of live module ids, and restores the `cluster/module_uuid` invariant that `list-backups` now relies on.

NethServer/dev#8094

## Test plan
- [x] Create a backup schedule including two app instances, run it
- [x] Remove one of the instances from the cluster
- [x] Verify `list-backups` / the Backup UI no longer shows the removed instance
- [x] Remove a whole node with backed-up instances and verify the same
- [x] On a cluster with pre-existing dangling backup references, run a core update and verify the migration cleans them up

### DIFF
--- a/core/imageroot/var/lib/nethserver/cluster/actions/list-backups/50list
+++ b/core/imageroot/var/lib/nethserver/cluster/actions/list-backups/50list
@@ -66,6 +66,9 @@ for bid, backup in backups.items():
     instances = rdb.hget(f"cluster/backup/{bid}", "instances") or ""
     for mid in instances.split():
         module_uuid = rdb.hget("cluster/module_uuid", mid)
+        if module_uuid is None:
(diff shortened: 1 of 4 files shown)


DavidePrincipi:
Tests are **VERIFIED**

gsanchietti (approved):

PR #1270, opened 2026-08-04
fix(core): reduce cluster-admin requests and handle 429

andre8244:
## Summary

Opening a cluster-admin page issues many requests at once — the app's own assets, then one round trip per card. That is more than the new rate limiter allows in a single burst, so some requests are rejected and the cards behind them never finish loading. What runs out is the limiter's allowance for simultaneous requests, not its limit on sustained traffic.

Six of the eleven fixes analysed in the issue, lettered as they are there:

| | Change | Where |
|---|---|---|
| **A** | Burst default `100` → `300`. `GLOBAL_RATE_LIMIT_AVERAGE` stays at `25`: the sustained rate is what bounds a flood, so over 60 s an attacker gains 12% (1800 vs 1600 requests) while a legitimate page load, needing ~150 at once, stops being rejected. Mirrored in `api-moduled`. | `configuration.go`, `api-moduled.go` |
| **B** | Return after a failed task-status request, instead of dereferencing an undefined response and throwing a `TypeError`. | `notification.js` |
| **C** | Treat `429` as temporary in the agent's retry list. Nested tasks reach the leader through the same rate-limited api-server, so a throttled call aborted the whole task instead of backing off. | `apiclient.py` |
| **D** | Coalesce concurrent task-context requests. `websocket.js` dispatches several progress messages per task without awaiting, so each reached the cache-miss branch and issued its own `GET …/context` — measured 34 requests for 13 tasks, one fetched 4×. | `notification.js` |
| **E** | Send `Retry-After: 1` on a throttled response. | `middleware.go`, `api-moduled.go` |
| **F** | Back off on `429` in the task poll timer instead of re-arming immediately. | `notification.js` |

Also: `RateLimiter` clamps a burst below `1`. A zero-capacity bucket can never hold the whole token a request needs, so `*_BURST=0` rejected everything — an easy misreading, since `0` does disable the limiter on the `_AVERAGE` variable.

## Scope and caveats

- **This does not make the UI recover from a `429` it receives.** A card whose status request fails still needs the request retried, which is **H** and deferred. In particular **B** stops an unhandled promise rejection but does not release a component waiting on `*-completed`; that event is unreachable either way once the request has failed.
- **Deferred:** **H** (429 retry in the axios interceptor), **I** (webpack prefetch, one line, −47 requests, and the only lever that helps a cold load), **J**, **K**.
- **Static assets are deliberately not exempted** from the limiter: `gzip.Gzip` is registered before the static handler and recompresses every asset on every request, so exempting that path would leave an unauthenticated CPU-amplifying endpoint unprotected.
- **`Cache-Control` on static assets was tried and dropped.** `http.FileServer` sets `Last-Modified`, so heuristic freshness already caches them — for more than the `max-age=3600` it added, on any asset older than ten hours. Setting `no-cache` on `index.html` is worth doing, but it targets a pre-existing hazard unrelated to throttling and needs its own change.
- **No automated coverage.** The repository has no Go or UI unit tests and adding the first ones was out of scope.

## Related issue

NethServer/dev#8124

### DIFF
--- a/core/api-moduled/AGENTS.md
+++ b/core/api-moduled/AGENTS.md
@@ -82,7 +82,12 @@ All via `AMLD_` prefixed environment variables (uses Viper):
 | `AMLD_ID_KEY` | `uid` | JWT identity claim key |
 | `AMLD_EXPORT_ENV` | — | Extra env vars for handlers |
 | `AMLD_RATE_LIMIT_AVERAGE` | `25` | Per-IP requests/sec across all routes; `0` disables |
-| `AMLD_RATE_LIMIT_BURST` | `100` | Burst allowance above the average |
+| `AMLD_RATE_LIMIT_BURST` | `300` | Burst allowance above the average |
+
+The sustained average is what bounds a flood; the burst only sets how many
+requests may arrive at once, and is sized for a legitimate UI page load. A burst
+below `1` is clamped to `1` by `RateLimiter`; use `AMLD_RATE_LIMIT_AVERAGE=0` to
+disable the limiter.
 
 ## Key differences from api-server
 
--- a/core/api-moduled/README.md
+++ b/core/api-moduled/README.md
@@ -37,7 +37,14 @@ Configuration parameters can be passed as environment variables.
   across all API routes -- default `25`; set to `0` to disable rate limiting
 
 - `AMLD_RATE_LIMIT_BURST`, burst allowance above the average before requests
-  are rejected with HTTP 429 -- default `100`
+  are rejected with HTTP 429 -- default `300`
+
+The sustained average is what bounds a flood; the burst only sets how many
+requests may arrive at once, and is therefore sized for a legitimate UI page
+load.
+
+A burst below `1` is raised to `1`, because a zero-capacity token bucket would
+reject every request; use `AMLD_RATE_LIMIT_AVERAGE=0` to disable the limiter.
 
 ## How to implement the API
 
--- a/core/api-moduled/api-moduled.go
+++ b/core/api-moduled/api-moduled.go
@@ -55,6 +55,12 @@ type rateLimiterVisitor struct {
 // (BodyLimit bounds one request's body, not how many requests run at once).
 // rps is the sustained rate and burst the number of requests allowed instantly.
 func RateLimiter(rps rate.Limit, burst int) gin.HandlerFunc {
+	// A zero-capacity bucket can never accumulate the whole token a request
+	// needs, so it would reject everything: "0 disables" only holds for rps.
+	if burst < 1 {
+		burst = 1
+	}
+
 	visitors := make(map[string]*rateLimiterVisitor)
 	var mu sync.Mutex
 
@@ -85,6 +91,10 @@ func RateLimiter(rps rate.Limit, burst int) gin.HandlerFunc {
 		mu.Unlock()
 
 		if !limiter.Allow() {
+			// The rate is configured as a positive integer, so a token always
+			// refills within a second and Retry-After, which is expressed in
+			// whole seconds, is always 1.
+			c.Header("Retry-After", "1")
 			c.JSON(http.StatusTooManyRequests, gin.H{"code": http.StatusTooManyRequests, "message": "too many requests", "data": nil})
 			c.Abort()
 			return
@@ -118,7 +128,7 @@ func main() {
 	viper.SetDefault("jwt_realm", "api-moduled")
 	viper.SetDefault("export_env", "")
 	viper.SetDefault("rate_limit_average", 25)
-	viper.SetDefault("rate_limit_burst", 100)
+	viper.SetDefault("rate_limit_burst", 300)
 	viper.AutomaticEnv()
 
 	logger = log.New(os.Stderr, "", 0)
--- a/core/api-server/AGENTS.md
+++ b/core/api-server/AGENTS.md
@@ -83,7 +83,13 @@ All via environment variables: `LISTEN_ADDRESS`, `REDIS_ADDRESS`, `REDIS_USER`,
 `REDIS_PASSWORD`, `SECRET` (JWT key), `STATIC_PATH`, `AUDIT_FILE`,
 `SENSITIVE_LIST`, `ISSUER` (TOTP issuer name, default `NethServer`),
 `GLOBAL_RATE_LIMIT_AVERAGE` (per-IP requests/sec across all routes, default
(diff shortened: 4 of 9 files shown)


DavidePrincipi:
## Review

Direction is right, and **A** and **C** are clearly correct. Below: three findings, then a proposal to reduce this PR to the minimum set that closes the issue and move the rest to follow-ups.

### 1. Fix B does not do what the description says — it is cosmetic

`core/ui/src/mixins/notification.js:283`. In the pre-PR code the order is:

```js
if (err) { console.error(err); ...; this.createNotification(notification); }   // toast already shown
taskResult = statusResponse.data.data;                                        // TypeError here
```

The error notification is **already created before** the throw. And with the new `return`, the function still never reaches the `if (taskResult) { … }` block that emits the `<action>-completed` event — so a card waiting on `<action>-completed` stays in its loading state either way. The only behaviour change is "unhandled promise rejection in the console" → "clean return".

So the claim in the description — *"The UI then turned that transient 429 into a permanent failure: affected Status cards stayed in the loading state forever … **B** … this is what left cards loading forever"* — is not supported by the code. What actually unsticks the card is **H** (retry the status request on 429), which is deliberately deferred here.

Two ways out:

- add the retry right at this spot, which is the smallest possible fix for the reported symptom:

  ```js
  if (err?.response?.status === 429) {
    await new Promise(r => setTimeout(r, (Number(err.response.headers?.["retry-after"]) || 1) * 1000));
    return this.handleProgressTaskMessage(taskPath, taskId, payload);   // retry once via re-entry
  }
  ```

- or keep the plain `return` and reword both the description and the code comment to "avoids an unhandled `TypeError`", leaving the stuck-card fix to **H**.

### 2. `StaticCacheControl` adds a `stat(2)` to every request, including `/api`

It is a global `router.Use`, and `localFileSystem.Exists` stats `STATIC_PATH + <url path>` — which `static.Serve` then does again for the same request. Since the two middlewares are registered adjacently and already share the file system, collapsing them does the lookup once:

```go
fileserver := http.StripPrefix("/", http.FileServer(staticFileSystem))
router.Use(func(c *gin.Context) {
    if c.Request.Method != http.MethodGet && c.Request.Method != http.MethodHead ||
        !staticFileSystem.Exists("/", c.Request.URL.Path) {
        c.Next(); return
    }
    c.Header("Cache-Control", staticCacheControl(c.Request.URL.Path, maxAge))
    fileserver.ServeHTTP(c.Writer, c.Request)
    c.Abort()
})
```

~10 lines replacing two registrations, one stat per request instead of two, and the "cannot disagree with the static handler" property becomes structural instead of a comment.

### 3. The burst-must-not-be-0 hazard is documented in four files instead of guarded in one line

The same paragraph is now in `AGENTS.md` + `README.md` of both components. `if burst < 1 { burst = 1 }` inside each `RateLimiter` makes the footgun unreachable and lets the docs shrink to a sentence.

## Cheaper alternatives for the remaining fixes

**E — 36 duplicated lines can be one line.** `retryAfterSeconds` is copied verbatim into `middleware.go` and `api-moduled.go`, and for any sane configuration it can only return `1`: at the default 25 rps the missing fraction of a token refills in ≤ 40 ms, and `average` would have to drop below 1 rps before `ceil((1-tokens)/rps)` exceeds one second. The `Retry-After: 1` observed on the live node *is* the computed value, not a coincidence. So `c.Header("Retry-After", "1")` is equivalent; if sub-1-rps configs should stay honest, `strconv.Itoa(max(1, int(math.Ceil(1/float64(rps)))))` needs neither the helper nor `limiter.Tokens()`. Fix **F**'s `taskProgressPollPeriod * 4` fallback already covers a missing header.

**G — `staticMustRevalidate` can be one regex test.** The `/apps/<id>` re-rooting, the `config/` case, the directory and `index.html` cases all exist only to compute "not cacheable", which is exactly the complement of "carries a content hash". Non-hashed files (`index.html`, `config/*`, `css/core.css`, `shortcuts.json`, `favicon.ico`) fail the hash test by construction, and directories have a hash-free basename:

```go
if maxAge > 0 && staticContentHash.MatchString(path.Base(c.Request.URL.Path)) {
    c.Header("Cache-Control", fmt.Sprintf("public, max-age=%d", maxAge))
} else {
    c.Header("Cache-Control", "no-cache")
}
```

Identical behaviour, ~45 lines and three edge-case explanations less. Whichever form lands, it deserves a small table test over the paths currently described in prose — none of this logic is covered by a test today.

**G — and is it worth landing here at all?** As the description notes, it only helps repeat loads, and the request-count drop is unmeasured because the instrumented browser runs with the HTTP cache disabled. **I** (webpack prefetch, one line, −47 requests) is strictly cheaper *and* helps the first load, which G cannot.

**D — 55 lines of promise coalescing to save ~21 of ~150 requests.** The root cause is that `websocket.js:62` dispatches `handleProgressTaskMessage` without awaiting. If serialising per connection is acceptable, an `await` there (making `handleTaskMessage`/`onMessage` async) removes the concurrency that creates the duplicates, and covers any other interleaving hazard in that handler rather than just this one. Either way, with **A** in place 21 requests is nowhere near the limit, so D is optimisation rather than fix.

## Proposal: ship the minimum set, split the rest

Minimum set that closes NethServer/dev#8124:

- **A** — burst `100` → `300` in both components (2 lines). This alone makes the reported failure disappear.
- **C** — `429` in `http_temporary_errors` (1 line).
- **E** — `Retry-After` as the one-liner above, in both components.
- **F** — the poll back-off on `429`, keeping the `Retry-After` header it now gets.
- plus one of the two options under finding 1, so the stuck-card claim matches the code.
- plus the `burst < 1` clamp from finding 3, with the docs trimmed accordingly.

That is roughly 30 lines of behaviour change instead of ~340, all of it directly on the reported failure path, and it can be reviewed and released without waiting for a measurement.

Follow-ups, each independently reviewable:

1. **G** static asset caching — with the collapsed middleware, the single-regex rule, a table test, and a repeat-load measurement in a browser with a live cache.
2. **I** webpack prefetch tuning — the cheapest remaining lever, and the only one that helps a cold load.
3. **H** `429` retry in the axios interceptor — absorbs residual throttling UI-wide, and supersedes the local retry if the finding-1 option above is taken.
4. **D** task-context coalescing — ideally as the `await` in `websocket.js` rather than a second cache layer.
5. **J** / **K** as already recorded in the issue.

<sub>Assisted-by: Claude Code:claude-opus-5</sub>

DavidePrincipi (approved):

gsanchietti (approved):

PR #1272, opened 2026-08-07
Add regexp mode to system logs search

stephdl:
Adds a "Regular expression" toggle to the System Logs search bar. When enabled, the filter is applied with the LogQL `|~` operator instead of the substring `|=`.

Backslashes are now escaped along with quotes for both operators, so a filter carrying either no longer produces an unparsable query and a misleading "No log found".

Invalid patterns are rejected client-side, and again server-side with Go regexp (RE2, the engine Loki runs). logcli failures now reach the UI as an error notification: until now a broken query was indistinguishable from an empty result.

Also from the Figma mockup:

- a feedback line showing the number of retrieved lines and the interval they cover
- collapse-filters and layout controls moved out of the kebab menu into the page toolbar
- filter field widths aligned

Refs NethServer/dev#8032
Design: NethServer/dev#7974

## Testing

- toggle off, filter `redis`: results unchanged
- toggle on, filter `redis|traefik` and `^2026-05`
- toggle on, filter `[unclosed`: error notification, no empty state
- toggle off, filter containing a backslash: results instead of a silent empty
- each of the above in both Dump and Follow mode
- entity selection and date-range filtering still narrow the results
- CLI: `api-server-logs logs -s 'err(or)?' -r -m dump -l 50`


core/api-server/README.md
DavidePrincipi, line 1: core/api-server/README.md:391 — the CLI --help block wasn't updated. It's missing both -r, --regexp (this PR) and -i, --instance (pre-existing). Since the PR already documents instance for the socket payload, adding both here would be consistent. No shorthand collision on -r/-i.

core/api-server/socket/action.go
```diff
@@ -291,6 +296,57 @@ func Action(socketAction models.SocketAction, s *melody.Session, wg *sync.WaitGr
...
 }
 
+// Reuses logs-start so the frontend one-shot listener is released.
+func writeLogsError(s *melody.Session, id string, message string, wg *sync.WaitGroup) {
+	utils.LogError(errors.New("[SOCKET] logs query failed: " + message))
+
+	if s != nil {
+		writeSocketResponse(s, "logs-start", gin.H{"id": id, "pid": "", "message": "", "error": message})
+	} else {
+		fmt.Fprintln(os.Stderr, message)
```
DavidePrincipi, line 314: The message prints twice. It was already printed by line 301.

```suggestion
```

core/api-server/socket/action.go
```diff
@@ -143,29 +142,13 @@ func Action(socketAction models.SocketAction, s *melody.Session, wg *sync.WaitGr
...
-			streamSelector = `{module_id="` + logsAction.EntityName + `"}`
-		}
-
-		logqlPipeline = ` | json syslog_id="SYSLOG_IDENTIFIER", message="MESSAGE" | line_format "[{{.node_id}}:{{.module_id}}:{{.syslog_id}}] {{.message}}"`
-
 		// Compose and append the query strings to logcli arguments
-		args = append(args, streamSelector+logqlPipeline+filter)
+		query, errQuery := buildLogqlQuery(logsAction)
+		if errQuery != nil {
+			writeLogsError(s, logsAction.Id, errQuery.Error(), wg)
```
DavidePrincipi: Please do not log the error here. An invalid user regexp is a typo, not a server fault; it now lands in the api-server stderr/journal.

```suggestion
```

core/api-server/socket/action.go
```diff
@@ -224,6 +212,18 @@ func Action(socketAction models.SocketAction, s *melody.Session, wg *sync.WaitGr
 				// use Wait to avoid defunct process when killed
 				err = cmd.Wait()
 				if err != nil {
+					// logs-stop kills the process: that exit is expected
+					if exitErr, isExitErr := err.(*exec.ExitError); isExitErr {
+						if status, isStatus := exitErr.Sys().(syscall.WaitStatus); isStatus && status.Signaled() {
+							return
```
DavidePrincipi, line 228: Log error on unexpected signals (e.g. OOM killed)

```suggestion
							if status.Signal() == syscall.SIGTERM {
								return
							}
```

DavidePrincipi (commented):
The backend part seems ok, I'm proposing just a few little tweaks.

core/ui/src/components/system-logs/LogSearch.vue
```diff
@@ -128,52 +144,64 @@
...
         </cv-row>
         <cv-row>
           <cv-column :md="verticalLayout ? 8 : 4">
             <cv-select
               :label="$t('system_logs.timezone')"
-              class="mg-bottom-md"
+              class="mg-bottom-lg"
               v-model="internalTimezone"
             >
               <cv-select-option disabled selected hidden>{{
```
andre8244, line 204: Timezone field should not be displayed in follow mode

core/ui/src/components/system-logs/LogSearch.vue
```diff
@@ -279,15 +311,14 @@
...
+      <cv-row v-if="logsError">
         <cv-column>
-          <cv-link @click="toggleFilters" class="toggle-filters">
-            {{
-              filtersShown
-                ? $t("system_logs.collapse_filters")
-                : $t("system_logs.expand_filters")
-            }}
-          </cv-link>
+          <NsInlineNotification
```
andre8244: Let's move this error notification below the Search button

core/ui/src/components/system-logs/LogSearch.vue
```diff
@@ -897,7 +1042,21 @@ export default {
       // signal LogOutput
       this.$root.$emit(`logsUpdated-${this.searchId}`);
     },
+    onLogsQueryError(message) {
+      this.loading.logs = false;
+      this.loading.stopFollowing = false;
+      this.isFollowing = false;
+      this.pid = "";
+      this.noLogsFound = false;
+      this.logsError = message;
```
andre8244: The regex error is not internationalized 🤔

core/ui/public/i18n/en/translation.json
```diff
@@ -1178,15 +1178,24 @@
...
+        "time_format_helper": "24-hour",
         "start": "Start",
         "end": "End",
         "context": "Context",
         "context_cluster": "Cluster",
         "context_node": "Node",
         "context_module": "Application",
         "search_query": "Search query",
+        "regexp": "Regular expression",
+        "regexp_helper": "RE2 syntax is accepted. Hint: (?i) ignores case, | matches any of the words",
```
```suggestion
        "regexp_helper": "RE2 syntax is accepted. Hint: '(?i)' ignores case, '|' matches any of the words",
```

andre8244:
It looks like some patterns (e.g. `()`, `(.)`...) overload the browser memory and freeze the web page

core/api-server/README.md
stephdl: Done in 7278244b, and I copied the actual binary output instead of hand-writing it: the column alignment and the -z/-t order were wrong in the file too.

core/api-server/socket/action.go
stephdl: Done in 1890449e. I had first moved the print into a shared helper, which kept the duplicate: utils.LogError already writes to stderr. Each path now prints once, the server fault through utils.LogError and the user error through its own fmt.Fprintln.

core/api-server/socket/action.go
stephdl: Done in 4b839d76. The invalid regexp goes through writeLogsQueryError, which does not log: it only sends the message back with an error_code the frontend turns into a translated wording. Checked on a test node, journalctl -u api-server stays empty after a failed pattern.

core/api-server/socket/action.go
stephdl: Done in 926d5658, your suggestion as is: only SIGTERM returns, so an OOM kill is reported.

core/ui/src/components/system-logs/LogSearch.vue
stephdl: Done. The error banner is translated (title and sentence), and the engine detail is appended as it comes: "The search pattern is not valid RE2 syntax: error parsing regexp: missing closing ): `(`". That tail is produced by Go's regexp package and names the offending token, so it stays in English rather than being dropped.

core/ui/src/components/system-logs/LogSearch.vue
stephdl: Done in 7e2af902, the notification now renders after the toolbar row. It is also hidden when the search returned no lines, where it read as a failure report on an empty result.

core/ui/public/i18n/en/translation.json
stephdl: Applied. The string then had to change once more: regexp mode no longer highlights matches, so it now reads "RE2 syntax is accepted, matches are not highlighted. Hint: '(?i)' ignores case, '|' matches any of the words". Tell me if you prefer that second sentence elsewhere.

stephdl:
The freeze is gone, and not by containing it: regexp mode no longer highlights matches at all, so no pattern written by the user is ever compiled by the browser engine. Substring mode keeps the highlighter it had before this branch, unchanged.

Measured on a test node, 500 lines, worst main-thread stall per search:

| pattern | marks | stall |
|---|---|---|
| `()`, `(.)`, `.*`, `\b`, `^`, `$` | 0 | <= 24 ms |
| `(.*)*x`, `(a+)+$`, `((ab)*)*c` | 0 | <= 29 ms |
| `(?P<lvl>...)`, `\Qsystemd\E`, `[[:alpha:]]+` | 0 | <= 32 ms |
| substring `e`, 2000 lines | 16106 | 681 ms |

~~The helper text under the field says matches are not highlighted when the toggle is on.~~

core/ui/src/components/system-logs/LogSearch.vue
stephdl: Done in 2d77e1f8, the field and its entry in the collapsed filter summary are both hidden while following.

Digging further, hiding it was not enough: the timezone applies to follow output too (logcli -z UTC prints 07:23:02Z, -z Europe/Rome prints 10:01:01+02:00 on the same lines), so a UTC picked for an earlier dump search kept driving live timestamps with no visible control. Follow mode now always uses the browser timezone, 975657f8. Verified: dump in UTC shows 2026-08-26T07:11:48Z, switching to follow shows 2026-08-26T10:05:36+02:00.

core/ui/public/i18n/en/translation.json
stephdl: back to your proposal

andre8244 (approved):
UI part LGTM

DavidePrincipi (commented):
The last commit removes the inline notification that displays backend error details. Since this part required a change of the websocket protocol, I preferred to make a step back and report backend errors in the text area itself.

I know it is less effective for UX, but the changes introduced by the feature (which should be rarely used) should be kept small.

I'm requesting Ande's review again.

andre8244 (approved):

DavidePrincipi (approved):
Filed upstream bug https://github.com/grafana/loki/issues/24249 about tail mode hangs in some cases.
NS8 bug https://github.com/NethServer/dev/issues/8140 for core 3.22+

PR #1280, opened 2026-08-27
  refactor(system-logs): print query errors as logs

DavidePrincipi:
A failed logs query is already described by logcli and Loki, and their
  wording is log output: the browser can print it in the text area
  instead of the backend growing a parallel error channel to carry it.
  Errors stop travelling next to the logs and start travelling as logs.

  - logs-start carries the failure in message, dropping the error and
    error_code attributes that every consumer had to special-case
  - writeLogsQueryError is gone, and writeLogsError no longer takes the
    WaitGroup of the CLI, which the exit already ends
  - buildLogqlQuery no longer compiles the pattern: Loki rejects a bad
    regexp itself, so invalid_regexp had nothing left to report
  - the tail goroutine sends logs-stop whenever it returns, which is how
    the frontend already leaves the follow state          
  - the frontend drops the error banner and its two i18n keys; a lost
    websocket is appended to the output like a server-side failure


stephdl:
Thanks, this is much better. `go build`, `go vet` and `yarn lint` are all clean on my side, and the RE2 error now lands in the output area exactly like your `logcli` example.

One thing I noticed while reading the `tail` goroutine: the deferred `logs-stop` fires on every exit path, including the `SIGTERM` one. But that path is precisely the one where the `logs-stop` action handler has already sent a `logs-stop` response (`action.go:81`). So a user-initiated stop puts two `logs-stop` on the wire.

Today the second one is unowned, because `$once` consumed the first. But the search panel keeps its `searchId` across searches. If a new follow search registers its listener before the stray arrives, the fresh follow is torn down in the UI while `logcli` keeps streaming. The window is microseconds, so it is latent rather than live — but the `SIGTERM` branch already knows the exit was expected, so the guard is cheap:

```go
stoppedByUser := false
if s != nil {
	defer func() {
		if !stoppedByUser {
			writeSocketResponse(s, "logs-stop", gin.H{"id": logsAction.Id, "pid": pid, "message": "logs follow stopped"})
		}
	}()
}
```

and set `stoppedByUser = true` in the `SIGTERM` branch before its `return`. Is it worth it, or do you prefer to keep the defer unconditional and let the duplicate be?

Also, minor and I do not plan to change it: `resultsFeedback` counts the error line, so a failed dump reads "Showing 1 of 500 lines" under the error text. Suppressing that would need the per-error state this patch removes.

stephdl:
I merged the patch into `sdl-8032` and tested it on a real node. `go build`, `go vet` and `yarn lint` are clean, the new binary and the new UI assets are in place.

**`dump` mode works exactly as you designed it.** The output area shows the `logcli` text verbatim, and the CLI does the same:

```
# api-server-logs logs -m dump -r -s 'GET(' -l 5
exit=1
2026/08/27 18:31:40 error response from server: parse error : stage '|~ "GET("' : error parsing regexp: missing closing ): `GET(` (<nil>) attempts remaining: 0
2026/08/27 18:31:40 Query failed: run out of attempts while querying the server
```

**`tail` mode does not.** With an invalid pattern, `logcli query -t` never exits and never writes anything — not on stdout, not on stderr. I checked it outside our code, so this is `logcli`, not us:

```
# timeout 15 logcli query -q --no-labels -t --forward '{node_id=~".+"} |~ "GET("'
exit=124        <- killed by timeout, logcli was still running
stderr: empty
stdout: empty
```

Loki does log the rejection on its side:

```
level=error caller=tail.go:230 component=tail-querier msg="Error receiving response from grpc tail client" err="rpc error: code = Code(400) desc = parse error : stage '|~ \"GET(\"' : error parsing regexp: missing closing ): `GET(`"
```

but the gRPC tail client swallows it and keeps reconnecting. So in follow mode `cmd.Wait()` never returns: no error `logs-start`, no deferred `logs-stop`, and the frontend sits in the follow state with an empty output area forever. That is the case the RE2 pre-check used to cover. "Let `logcli` report the error" holds for `dump` only.

Three ways out, your call:

1. Put back `regexp.Compile` and nothing else. No error code, no template, no i18n key: the Go error text is sent as `message` on `logs-start`, so it prints in the output area like any other log text, same as the `dump` path. About five lines of Go, and it keeps the shape you asked for.
2. Make `logcli` surface it. I do not see a clean lever — `--retries=0` does not apply here, the loop is inside the gRPC tail client, not in the query retry logic.
3. Accept that follow mode hangs on a bad pattern and document it.

I lean towards 1, but the patch is yours.

PR #1292, opened 2026-09-03
fix(ui): prevent foreign Vue observers from freezing the cluster-admin SPA

edospadoni:
## Summary

Fixes the whole-SPA browser freeze triggered by module UI pages (see the linked issue for the full analysis and captured stacks).

Module UIs run in same-origin iframes with their own bundled copy of Vue and receive the live core root instance via `window.parent.core`; several module UIs keep it in their reactive state. Vue recognizes "do not observe" values only with realm- and version-local checks (Vue 2.6 skips `_isVue`, Vue 2.7 skips `__v_skip`), so a module bundling Vue 2.7 deep-observes a Vue 2.6 core component tree and rewrites its internals into foreign reactive getters; from then on every core render triggers a full foreign observation walk, freezing the whole tab until reload.

Changes:

- `core/ui/src/main.js`: global mixin marking every core instance with `__v_skip = true`, honored by Vue 2.7 and Vue 3 observers regardless of realm. Vue 2.6 does not read the flag and Vue 2.7 sets it natively on its own instances, so this is a cross-version guarantee with no behavior change for the core itself. This is the change that fixes the freeze.
- `core/ui/src/store/index.js` and `core/ui/src/mixins/notification.js`: hardening of the opposite direction. Task contexts and results were inserted into the reactive notifications store by reference while the same objects are handed to module UIs through the `$root.$emit` bridge; a new `toPlainCopy` helper (cycle-safe deep plain copy; keeps functions such as notification action callbacks and non-plain leaves such as `Date`; discards Vue instances and VNodes with duck-typed checks) is applied in the `createNotification`/`updateNotification` mutations and in the subtask insertion path, which mutates the store directly.

## Related issue

NethServer/dev#8147

## How to test

1. Cluster with a Vue 2.6 based core build and a module UI bundling Vue 2.7 that stores `window.parent.core` in its Vuex state (e.g. current NethVoice UI).
2. Without the fix, navigating the module pages (settings/status/hotel) freezes the tab within a few page loads (reproduced deterministically via headless browser + CDP stack capture).
3. With the fix, the freeze no longer reproduces (verified empirically: 30 consecutive module page transitions on the affected test cluster, plus manual navigation and settings save).
4. Verify regular notification behavior still works: toast on task completion, notification drawer, task error details modal, the "Reload" action of the websocket-disconnected notification, subtask progress rendering.

## Dependencies

None.

### DIFF
--- a/core/ui/src/main.js
+++ b/core/ui/src/main.js
@@ -95,6 +95,21 @@ Vue.config.productionTip = false;
 import VueI18n from "vue-i18n";
 import { loadLanguage } from "./i18n";
 
+// Module UIs run in same-origin iframes with their own copy of Vue and can
+// keep references to core component instances in their reactive state (e.g.
+// `window.parent.core` ends up in a module Vuex store). A foreign Vue >= 2.7
+// observer cannot recognize instances of another Vue copy (its checks are
+// realm-local) and would deep-observe the whole core component tree,
+// rewriting its internals into foreign reactive getters and freezing the
+// entire SPA on the next render. Vue 2.7 and Vue 3 observers skip any value
+// marked with __v_skip, so mark every core instance explicitly (Vue 2.6 sets
+// only _isVue, which foreign Vue 2.7 observers do not check).
+Vue.mixin({
+  beforeCreate() {
+    this.__v_skip = true;
+  },
+});
+
 loadI18n();
 
 async function loadI18n() {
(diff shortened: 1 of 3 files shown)


edospadoni:
Closing: the affected core image was an outdated build of feat/backup-my-cutover predating the Vue 2.7 upgrade (152d8493). The branch has been rebased onto main and the image rebuilt from it, which removes the Vue 2.6/2.7 mismatch described in NethServer/dev#8147, so this change is not needed.

PR #1295, opened 2026-09-04
Document Redis options in the Python agent

Amygos:
Clarify that `agent.redis_connect()` forwards additional keyword
arguments to the installed Redis Python library's `redis.Redis()`
constructor. Module authors can therefore configure supported client
options such as socket timeouts, retries, TLS, database selection, and
response decoding without dedicated agent API parameters.

The module developer guide now demonstrates both general connection
options and `decode_responses=False` for callers that need raw Redis bulk
strings. The function documentation and inline comment also make this
existing behavior explicit.


Amygos:
The autogenerated tests have been removed before marking the PR ready to review.

DavidePrincipi (changes requested):
This PR leaves the previous behavior unchanged. Even if it improves code readability, It is actually a no-op.

DavidePrincipi (approved):

PR #1296, opened 2026-09-04
fix(ldapclient): handle AD attributes hidden from bind account

DavidePrincipi:
Listing users from a remote AD crashes when one of the accounts has fields Windows won't let the bind account read — this happens with protected accounts like `krbtgt_AzureAD` (created by Azure AD Kerberos). AD returns those fields as empty instead of just leaving them out, and the code wasn't ready for that.

Also fixes a related issue: when a user's password was never set, the list-users action showed a bogus expiration date far in the past (-11636697600) instead of a neutral value.

NethServer/dev#8151
NethServer/dev#8153

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/agent/ldapclient/ad.py
+++ b/core/imageroot/usr/local/agent/pypkg/agent/ldapclient/ad.py
@@ -117,11 +117,14 @@ class LdapclientAd(LdapclientBase):
                 "description": description,
             }
 
+        # userAccountControl may be an empty list if the bind account is not
+        # authorized to read it (e.g. remote Microsoft AD)
+        uac = entry['attributes']['userAccountControl'] or 0
         return {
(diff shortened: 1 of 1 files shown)


stephdl (approved):
The fix looks good to me, and the crash it removes is real.

One question: `password_expiration = 0` is a new third value, but
`validate-output.json` still describes the field as "UNIX timestamp when the
password will expire, or -1 if it never expires". `0` is a valid timestamp
(1970-01-01), so a consumer that formats it as a date shows "expired since
1970". The core UI is safe, but `list-domain-users` output is consumed by
ns8-* modules too.

Document `0` in the description, or keep `-1` and use `must_change` as the
discriminant?

stephdl (approved):
Thanks, `-1` with `must_change` as the discriminant is exactly what I was after. It matches the existing `validate-output.json` description, so no doc change is needed. Good for me.

PR #1299, opened 2026-09-08
feat(firewall): list rich rules

gsanchietti:
This PR adds the `list-rich-rules` node action and the `agent.list_rich_rules()` helper, granted to the `fwadm` role.
Mimic `add-rich-rules` and `remove-rich-rules behavior.

Issue: https://github.com/NethServer/dev/issues/8157

**How to test**

Update an existing machine:

```
api-cli run update-core --data '{"core_url":"ghcr.io/nethserver/core:list_rich_rules","nodes":[1],"force":true}'
```

Add some rules, then list them:

```
api-cli run node/1/add-rich-rules --data '{"rich_rules": ["rule family=ipv4 forward-port port=5060 protocol=udp to-port=5060", "rule family=ipv4 source address=1.2.3.4 accept"]}'
api-cli run node/1/list-rich-rules --data '{}'
```

The output must match `firewall-cmd --permanent --list-rich-rules`. Remove a rule and list again: it must be gone. With no rules left, the action returns `{"rich_rules": []}` and exits 0.

Check the permission reached the role on an upgraded cluster:

```
redis-cli smembers node/1/roles/fwadm
```

It must contain `list-rich-rules`.


DavidePrincipi (approved):
Since this new function can be invoked by app migration code, I added `raise_on_error` to distinguish an empty rule list from an hard error.

gsanchietti:
> Since this new function can be invoked by app migration code, I added `raise_on_error` to distinguish an empty rule list from an hard error.

I did not want it by choice: I prefer an empty list.
But since default behavior is still an empty list, ok for me.

PR #1302, opened 2026-09-09
perf(cluster): parallelize and cache list-modules repository fetches

DavidePrincipi:
This is the other approach discussed for NethServer/dev#8161 (the resilience alternative, as opposed to the local-only `offline` flag in #1301): it reduces the Applications/Software Center page stall caused by unreachable repositories without changing `list-modules`' output shape or contract.

- Fetch every enabled repository concurrently with a thread pool instead of sequentially, so one unreachable repository no longer adds its full connect/read timeout on top of every other request. Benefits are measurable if more than one repository is enabled. For example, with three enabled repositories A, B, C the timeout is capped to Tmax(A, B, C) instead of T(A) + T(B) + T(C). Notes:
  1. In Python Requests `.get()` call,  T(X) -- the timeout of each repo -- is a function of IP addresses returned by DNS resolution. It's given by 4s x ips(X), where ips(X) is the number of IP addresses, like `getent ahosts X`.   
  2. The base 4s connect timeout value is raised to 10s for nightly updates.
- Cache a failed repository fetch as a short-lived (60s) error marker on the same `repository_cache` Redis key used for successful results, so a burst of near-simultaneous `list-modules` calls doesn't each pay the same timeout again.
- Fix fallback logo display in Applications page

See also https://github.com/NethServer/dev/issues/8088

**Test plan**

- [ ] With all repositories reachable, `list-modules` output is unchanged
- [ ] With one or more repositories unreachable, `list-modules` returns in roughly the time of the slowest single repository timeout, not the sum of all of them
- [ ] A second `list-modules` call within 60s of a failed fetch does not re-attempt the unreachable repository
- [ ] A repository that fails once and then recovers is not stuck reporting empty results after succeeding
- [ ] The application logo is displayed correctly for installed applications

### DIFF
--- a/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
+++ b/core/imageroot/usr/local/agent/pypkg/cluster/modules.py
@@ -136,6 +136,12 @@ def _get_http_session():
 def _list_repository_modules(rdb, repository_name, repository_url):
     cache_key = f'cluster/repository_cache/{repository_name}'
     hcache = rdb.hgetall(cache_key)
+    if hcache.get('error'):
+        # A request already failed for this repository a short time ago
+        # (see below): skip the remote query instead of paying its
+        # timeout again, e.g. on the next list-modules call right after
+        # this one.
+        return []
     repo_view = get_repo_view()
     if not hcache or hcache.get('repo_view') != repo_view:
         url = _urljoin(repository_url, "repodata.json")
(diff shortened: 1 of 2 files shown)


andre8244 (approved):
LGTM, just a little inaccuracy in the **Test plan**:

> A second list-modules call within 10s of a failed fetch does not re-attempt the unreachable repository

IIUC it should be 60s, not 10s

stephdl (approved):
Approving.

One small thing for later: `hset` then `expire` are two commands, so a Redis hiccup between them leaves the marker with no TTL and the repository stays hidden until a manual `DEL`. A single `rdb.set(key, "1", ex=60)` on a dedicated key would close that. Not blocking.

DavidePrincipi:
:no_good: ex= argument does not exist!


```
>>> rdb.hset('mykey', {"a":"2","b":"3"}, ex=10)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: HashCommands.hset() got an unexpected keyword argument 'ex'
```

PR #1306, opened 2026-09-15
docs(user_domains): fix headings, clarify mail lookup

DavidePrincipi:
## Summary
- Fix heading levels under "Password expiration warning" so subsections nest correctly
- Clarify how `notify-password-warning` resolves a user's destination address: LDAP `mail` attribute first, then a Mail application whose domain matches `user_domain` as fallback, noting that fallback only works if that Mail app is also the cluster's SMTP server for notifications

## Test plan
- [ ] Render docs locally and verify heading hierarchy/TOC
- [ ] Verify wording matches current `notify-password-warning` script behavior


gsanchietti (approved):
❤️

PR #1307, opened 2026-09-15
Add phone extension field for users

DavidePrincipi:
## Summary
- Read the phone extension from the LDAP `telephoneNumber` attribute (used by both AD and RFC2307/OpenLDAP domains)
- Add an optional "Phone extension" box to the user form, next to Email

**Update**: the phone extension used to be read from AD's `ipPhone` attribute; the requirement changed to use `telephoneNumber` for AD too, aligning it with RFC2307/OpenLDAP.

NethServer/dev#8027

image

Input validation error:

image

Table view with additional column:

image

## Test plan
- [x] Open a user's edit form and check the new field shows up - Requires openldap and samba images from branch feat-8027: https://github.com/NethServer/ns8-openldap/pull/160 https://github.com/NethServer/ns8-samba/pull/242
- [ ] Check the tooltip text is clear

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1308, opened 2026-09-15
Add phone extension to CSV import and export

DavidePrincipi:
## Summary
- Add the phone extension column to the CSV file used to import and export users

**Note**: no code dependency on #1307, but both PRs add lines next to each other in the same translation file. Merge #1307 first to avoid a conflict.

**Rejected for backward compatibility**: this PR's strict CSV column-count check (`results.data[i].length !== expectedColumns`) rejects any CSV that isn't exactly the new column count, including existing 8-column exports/templates that predate this field. Needs rework to tolerate the old column count (or otherwise not hard-fail on it) before re-review.

The backend PRs (ns8-samba#244, ns8-openldap#161) are NOT affected by this and are safe to merge on their own: the extra `phone_extension` key they add to `export_users()` output is additive JSON, ignored by the current frontend's CSV export/import.

NethServer/dev#8027

## Test plan
- [ ] Export users to CSV, check the phone extension column shows up
- [ ] Import a CSV file with a phone extension column, check it is read correctly


andre8244 (approved):

DavidePrincipi:
Since this change breaks CSV format backward compatibility we will evaluate this change in a future round
https://mattermost.nethesis.it/nethesis/pl/exq4en5x5bnzux8k4oewh6xxho @andrea-marchionni

PR #1309, opened 2026-09-15
Node product DMI details

edospadoni:
## Summary

The `product` attribute of the node `get-facts` action response now carries the DMI details of the node:

```json
{
  "product": {
    "name": "NethBox Z1+",
    "version": "1.0",
    "uuid": "03000200-0400-0500-0006-000700080009",
    "manufacturer": "Nethesis",
    "board": "NethBox Z1+",
    "bios": {
      "version": "QHSW0201.V22",
      "vendor": "American Megatrends Inc."
    }
  }
}
```

Everything is nested under `product` to avoid repeating the manufacturer in a separate `board` object. Values come from the `ansible` setup module facts (`ansible_product_version`, `ansible_product_uuid`, `ansible_board_name`, `ansible_bios_*`) and are read with a default value, because DMI data is not available on every platform. The existing `name` and `manufacturer` attributes are unchanged.

The hardware `uuid` identifies the machine, so it is pseudonymized with the existing `agent.facts.pseudo_string()` helper: it is reported as is when the cluster has a subscription, and as a 16 characters digest, stable within the cluster, when it has not. This is the same behaviour of `inventory.anonymize()` in NethSecurity, see NethServer/nethsecurity#1898.

No change is needed in `print-phonehome`: `node_facts()` copies the action output as is, so the new attributes are forwarded to the `facts.nodes` section.

## Related issue

None, please link it if there is one.

## How to test

Run the action on any node of the cluster and check the `product` attribute:

```
api-cli run node/1/get-facts | jq '.[].product'
```

Verified on a two-node cluster (Rocky 9.8, ansible-core 2.19.1): with a subscription both nodes report their DMI values and their hardware uuid in clear text; without a subscription each node reports a distinct, stable digest in place of the uuid.


DavidePrincipi (approved):
LGTM

PR #1310, opened 2026-09-15
Add tooltip and align optional label for Mail field

DavidePrincipi:
## Summary
- Add a tooltip to the Email field explaining where the value is stored and how it's used
- Build the "(optional)" label the same way as Groups and Phone extension, instead of a separate string

Note: 

- this changes the field label text from  "Mail address (optional)" to "Email (optional)", since it now reuses the existing `domain_users.mail` string ("Email") instead of the old dedicated `mail_field` string ("Email (optional)").
- this updates the label everywhere domain_users.mail is used: the create/edit  form ("Email (optional)"), and both table headers (Users and Groups list, CSV import preview) that now  say "Email" instead of "Mail address".

image

image

image

- NethServer/dev#8027
- NethServer/dev#7545

## Test plan
- [ ] Open the create/edit user form, check the Email field tooltip text
- [ ] Check the Email field label now reads "Mail address (optional)"


core/ui/public/i18n/en/translation.json
```diff
@@ -1560,8 +1560,8 @@
         "group_with_same_name": "A group with this name already exists",
         "user_with_same_name": "A user with this name already exists",
         "display_name_string_lte": "Must be less than or equal to 256 characters",
-        "mail_field": "Email (optional)",
-        "mail": "Mail address",
+        "mail": "Email",
+        "mail_tooltip": "This value is saved in the LDAP 'mail' attribute and is used independently by applications, for example to send password expiration warnings, and configure the user's alias address in the Mail application",
```
```suggestion
        "mail_tooltip": "This value is saved in the LDAP 'mail' attribute and is used by applications, for example to send password expiration warnings, and configure the user's alias address in the Mail application",
```

andre8244 (approved):

PR #1312, opened 2026-09-16
Node network configuration facts

edospadoni:
## Summary

The node `get-facts` action response now describes the network setup of the node, the same data shown by the *Active network interfaces* table of the node firewall page:

```json
{
  "dns_servers": ["67.207.67.2", "67.207.67.3"],
  "network": {
    "configuration": {
      "eth0": {
        "name": "eth0",
        "type": "ethernet",
        "props": {
          "ipaddr": "164.92.145.170",
          "netmask": "255.255.240.0",
          "gateway": "164.92.144.1"
        },
        "addresses": [
          { "family": "inet", "address": "164.92.145.170", "prefix": "20" },
          { "family": "inet", "address": "10.18.0.32", "prefix": "16" }
        ]
      },
      "wg0": {
        "name": "wg0",
        "type": "tunnel",
        "props": {
          "ipaddr": "10.5.4.1",
          "netmask": "255.255.255.255",
          "gateway": ""
        },
        "addresses": [
          { "family": "inet", "address": "10.5.4.1", "prefix": "32" }
        ]
      }
    },
    "interface_counts": { "ethernet": 2, "bonds": 0, "bridges": 0, "vlans": 0, "tunnels": 1 },
    "route_info": { "count_ipv4_route": 5, "count_ipv6_route": 3 }
  }
}
```

Attribute names and nesting follow the `features.network` section of the NethSecurity inventory payload, so the two installations can be analyzed the same way. The section is attached to the node facts, because NS8 facts have no `features` object and the cluster reports one entry per node.

Values are taken from the `ansible` setup module facts, already collected by the action, so no further command is run except the route count. Name servers are read with `dns.resolver`, as the `print-nameservers` helper of ns8-samba does.

### Interfaces

- `configuration` lists the interfaces that are active and have at least one global address, as the node firewall page does.
- `addresses` lists every global address of the interface: unlike a NethSecurity appliance, an NS8 node often has more than one address on the same interface, and `props.ipaddr` alone would hide the others.
- `wg0`, the cluster VPN, is reported as well: it tells whether the node is part of the VPN mesh and which address it got. Interfaces created by the container runtime (`veth*`, `cni-*`, `podman*`, `docker*`) and the loopback are skipped instead: they are implementation details and they change at every module installation.
- `interface_counts` counts every active interface by type, including the ones with no address, like a bridge or a bond port.
- The NethSecurity `props.role` attribute is not reported: an NS8 node has no network roles, and any value would be arbitrary. The public or private scope of an interface can be told from its address, even when it is pseudonymized.

### Privacy

Every address, gateway and name server is pseudonymized by `agent.facts.pseudo_ip()` when the cluster has no subscription, exactly like the existing `default_ipv4` attribute: the public or private scope of the address is preserved, its value is not. Interface names, types, netmasks and counters are reported as they are.

## Related issue

None, please link it if there is one.

## How to test

Run the action on any node of the cluster:

```
api-cli run node/1/get-facts | jq '.[] | {dns_servers, network}'
```

The `configuration` attribute must list the same interfaces and addresses of the *Active network interfaces* table, in the node firewall page.

Verified on a two-node cluster (Rocky 9.8, ansible-core 2.19.1), running the action on the node itself, from the cluster leader and through `print-phonehome`: both nodes report their two Ethernet interfaces and the VPN one, with the public address, the private one and the VPN address. With the subscription disabled, every address is replaced by a stable pseudonym of the same scope. A `resolv.conf` that `dns.resolver` cannot parse logs a warning and leaves `dns_servers` empty, without failing the action.


core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
```diff
@@ -60,6 +62,119 @@ def _get_facts():
...
+    routes = _run(f"ip -{family} route show")
+    return len([route for route in routes.split("\n") if route])
+
+def _get_nameservers():
+    nameservers = []
+    try:
+        with open('/etc/resolv.conf', encoding='utf-8') as file:
+            for line in file:
+                if line.startswith('nameserver'):
+                    nameservers.append(agent.facts.pseudo_ip(line.split()[1]))
```
DavidePrincipi: A malformed nameserver-line may raise IndexError. Please handle that exception to skip the bogus line. Also check if pseudo_ip() may raise other exceptions to skip the line as well.

core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
```diff
@@ -60,6 +62,119 @@ def _get_facts():
...
+    elif '.' in iface:
+        return 'vlan'
+    return 'ethernet'
+
+def _interface_role(iface, address):
+    """Assign a network role to the interface, when its address tells one"""
+    if iface == 'wg0':
+        return 'gray' # cluster VPN
+    try:
+        if ipaddress.ip_address(address).is_private:
```
DavidePrincipi: What does "green" mean in NS8 context? A DigitalOcean VPS primary interface has a public IP address and would be assigned role `''` by this function. Is this behavior expected? 

Can we always return "green"?

core/imageroot/var/lib/nethserver/node/actions/get-facts/50get
DavidePrincipi: You may use `import dns.Resolver` library to parse resolv.conf 
Look at ns8-samba's `print-nameservers` helper for an example.

DavidePrincipi (approved):
LGTM

DavidePrincipi:
Refs https://github.com/NethServer/dev/issues/7942

